PR #1, opened 2019-05-06
Add refpool

piever:
I've added `refpool` (to return whenever available the "pool" of possible values indexed by the references) to generalize `t.pool` from PooledArrays.

As of know PooledArrays would need to define:

```julia
refarray(p::PooledArray) = p.refs
refvalue(p::PooledArray, x) = p.pool[x]
refpool(p::PooledArray) = p.pool
```

Maybe we could simplify things by changing `refvalue` to:

```julia
function refvalue(A::AbstractArray{T}, x)::T
    pool = refpool(A)
    pool === nothing ? x : pool[x]
end
```

but I don't know whether this would have an impact on performance on normal arrays.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -22,8 +22,9 @@ For a given array `A`, potentially return an optimized "ref array" representatio
 original array, which can allow for faster comparison and sorting.
 
 The default definition just returns the input array. This function is useful for custom
-array types which already store a "hashed"-like representation of elements where comparison
-can be much faster than the original scalar value, like pooled arrays.
+array types which already store a "hashed"-like representation of elements where testing
+equality or permuting elements in place can be much faster than the original scalar value,
+like pooled arrays.
 
 This generic function is owned by DataAPI.jl itself, which is the sole provider of the
 default definition.
@@ -35,7 +36,7 @@ refarray(A::AbstractArray) = A
     refvalue(A, x)
 
 For the *original* array `A`, and a "ref value" `x` taken from `refarray(A)`, return the
-appropriate *original* value.
+appropriate *original* value. `refvalue(A, refarray(A)[I...])` must be equal to `A[I...]`.
 
 By default, `refvalue(A, x)` returns `x` (since `refarray(A)` returns `A` by default).
 This allows recovering an original array element after operating on the "ref values".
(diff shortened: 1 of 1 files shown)


src/DataAPI.jl
```diff
@@ -46,6 +47,22 @@ default definition.
 function refvalue end
 refvalue(A::AbstractArray, x) = x
 
+"""
+    refpool(A)
+
+Whenever available, return an indexable object `pool` such that, given the *original* array `A` and
+a "ref value" `x` taken from `refarray(A)`, `pool[x]` is the appropriate *original* value. Return
+`nothing` i such object is not available. If `refpool(A)` is not `nothing`, then
```
quinnj: `if such`...

nalimilan:
Thanks! Do we need `refvalue` since `refpool(x)[i]` already gives access to the value?

piever:
One case where there is no `refpool` is something like `StructArray((u=u, v=v))` for example, with let's say `u` a `PooledArray` and `v` a normal `Array`. Then this can have as many unique values as the length of the array (as `v` is not pooled), but still `StructArray((u=refarray(u), v=refarray(v))` is a sensible `refarray` definition (and a very useful one as `StructArray` works much better if every field is isbits) and `refvalue(s, el) = (u=refvalue(s.u, el.u), v=refvalue(s.v, el.v))` is a useful definition.

One could of course decide that `refpool` returns a custom struct with a custom `getindex` instead of `refvalue`, but it would be a bit hard to define `keys` and `values` there. Maybe in that case it's better to have `refpool` return `nothing` (i.e. no obvious pool exists) and define `refvalue` as above.

nalimilan:
OK, makes sense.

nalimilan:
I realized when implementing this API for CategoricalArrays that the methods that the returned pool object must implement should probably be more explicitly listed here. Currently the docstring only says "return an indexable object". But if the only thing you can do is index it, it doesn't have any added value compared with calling `refvalue`. There should be a way to get the list of possible values. Should we require implementing `keys` or the iteration protocol?

bkamins:
it is probably natural to expect that `keys`, `values` and `pairs` are supported.

quinnj:
Yeah, the refpool could potentially be a NamedTuple, so I say we improve the docs to say that `keys`, `values`, and `pairs` need to be supported on the refpool.

nalimilan:
See https://github.com/JuliaData/DataAPI.jl/pull/24.

PR #7, opened 2019-07-27
Add All and Between from IndexedTables.jl

bkamins:
Fixes https://github.com/JuliaComputing/IndexedTables.jl/issues/261

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -73,4 +73,23 @@ definition.
 """
 function describe end
 
+"""
+    Between(first, last)
+Select the columns between `first` and `last` from a table.
+"""
+struct Between{T1 <: Union{Int, Symbol}, T2 <: Union{Int, Symbol}}
+    first::T1
+    last::T2
+end
+
+"""
+    All(cols...)
+Select the union of the selections in `cols`. If `cols == ()`, select all columns.
+"""
+struct All{T}
+    cols::T
+end
+
+All(args...) = All(args)
+
 end # module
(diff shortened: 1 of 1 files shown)


bkamins:
@piever + @quinnj Do you have an opinion on this change? (after the change IndexedTables.jl should be updated and I will update DataFrames.jl)

piever:
Sorry for the late reply! I'm in favor of this change and would like to take advantage of it to make the `All` constructor a bit stricter, i.e. leaving only one internal constructor `All(args...)` (same API as `union(args...)` from Base).

bkamins:
@piever Is my change in line with what you intended?

piever:
Yes, looks good to me!

bkamins:
@quinnj or @nalimilan - can please either of you approve this before merging.

quinnj:
This is fine this time, but consider adding tests next time; I realize it's mostly trivial, but it's at least good practice to ensure we don't have typos or anything.

bkamins:
Sure. Actually I have looked at the tests and concluded the policy is not to add tests for bare definitions (but now - looking at them again - I see that you do, but simply there is not much to test yet). I will add tests in a separate PR. Thank you.

src/DataAPI.jl
```diff
@@ -73,4 +73,22 @@ definition.
 """
 function describe end
 
+"""
+    Between(first, last)
+Select the columns between `first` and `last` from a table.
```
nalimilan: Sorry to be late, but I think the convention is to have an empty row after the signature, right?

bkamins: Right - I will add the fix (I copy pasted this from IndextedTables.jl, but now I see that it is correct there)

PR #12, opened 2019-10-02
Add levels

nalimilan:
Currently lives in Missings.jl, which isn't the most logical place. This will allow StatsModels to drop the dependency on Missings, while CategoricalArrays will continue adding a specialized method to it.

For reference about the behavior of `levels`, see discussion at https://github.com/JuliaLang/julia/pull/20800.

Cc: @kleinschmidt

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -74,6 +74,28 @@ definition.
 function describe end
 
 """
+    levels(x)
+
+Return a vector of unique values which occur or could occur in collection `x`,
+omitting `missing` even if present. Values are returned in the preferred order
+for the collection, with the result of [`sort`](@ref) as a default.
+
+Contrary to [`unique`](@ref), this function may return values which do not
+actually occur in the data, and does not preserve their order of appearance in `x`.
+"""
+function levels(x)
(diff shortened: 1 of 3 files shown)


src/DataAPI.jl
```diff
@@ -73,6 +73,28 @@ definition.
 """
 function describe end
 
+"""
+    levels(x)
+
+Return a vector of unique values which occur or could occur in collection `x`,
+omitting `missing` even if present. Values are returned in the preferred order
+for the collection, with the result of [`sort`](@ref) as a default.
```
bkamins: maybe add that `sort` happens only if the elements of the collection are sortable and that decision to try sorting is based on `eltype` of a collection and not the type of values?

E.g. `levels(Any[3,2,1])` returns `[3,2,1]`. Maybe this should also be added to tests?

nalimilan: Good catch. Let's drop the type check, that will make things more reliable.

bkamins: agreed

kleinschmidt:
Should also bump the version?

Issue #16, opened 2020-04-04
Rename `All` to `Union`, or to `Cols`?


pdeffebach:
It ocurred to me that a statement like `All(Not(:x), :y)` is a bit of an oxymoron. Also, if one were to construct a statement *outside* a `transform`-like function call you would see 

```
All(Not(:x), :y)
```

Which raises the question, `All` of what? 

Perhaps it could be `Union(Not(:x), :y)`. This reads better in my opinion.

piever:
One issue is that `All()` is used to select all columns, I think `Union()` would be very confusing in that sense (one would expect the empty set with `Union()`).

pdeffebach:
We could deprecate that in favor of `All(:)`. This is just an idea, I'm not wedded to it but think it might be helpful.

bkamins:
I agree that `Union` is a better name. Actually we had a bug in DataFrames.jl because `All()` is actually `All(:)` that was recently fixed. However, `Union` is not the best name as it is used by Base, so probably `All` should stay.

pdeffebach:
Thoughts on `Cols`? It can be thought of as a container for `Columns`. 

```
Cols(Not(:id:), r"^t", [:income, :consumption])
```

Makes it pretty clear that we are building a container of columns.

CameronBieganek:
I think that `Cols` is the best choice. The docstring for `All` is actually incorrect:

> All(cols...)
> Select the union of the selections in cols.

`All` creates a tuple of sets of columns, _not_ the union of the sets. For example,

```
All(:x, :y) != All(:y, :x)
All(:y, :) != All(:)
```

In DataFrames.jl the ordering produced by `All` is used to choose the order of the columns when indexing.

The actual behavior of `All` is this:

- Let `A`, `B`, and `C` be sets of columns.
- `All(A, B, C)` creates a tuple `t = (A, B\A, C\(A ∪ B))`.
- Then `df[:, All(A, B, C)]` becomes `df[:, vcat(t...)]`.

I think this behavior is much better captured by the `Cols` name. Then `Cols()` would select zero columns and `Cols(:)` would select every column.

Tagging @bkamins, since this would be a breaking change.

bkamins:
@piever - what do you think about it? I think that in DataFrames.jl it is not a problem to make this change now if it were preferred (but after 1.0 release this will be a problem), but this is a DataAPI.jl change that would have a larger impact.

CC @nalimilan

nalimilan:
[...]

PR #18, opened 2020-04-22
add Cols

bkamins:
Fixes https://github.com/JuliaData/DataAPI.jl/issues/16.

I think for now we should keep `All` to be non-breaking (and maybe just keep in in the long term - but this is a separate decision).


bkamins:
Do you know why tests have not triggered?

bkamins:
I have checked locally that the tests pass.

bkamins:
@nalimilan - are you OK with this? I would merge it and move forward in DataFrames.jl to include this functionality.

PR #25, opened 2020-10-21
Make levels docstring more precise

bkamins:
If `sort!` fails then (fortunately) no levels are lost, but it might reorder part of the vector before failing.


src/DataAPI.jl
```diff
@@ -85,6 +85,7 @@ function describe end
 Return a vector of unique values which occur or could occur in collection `x`,
 omitting `missing` even if present. Values are returned in the preferred order
 for the collection, with the result of [`sort`](@ref) as a default.
+If the collection is nod ordered then the order of levels is unspecified.
```
bkamins, line 88:
```suggestion
If the collection is not sortable then the order of levels is unspecified.
```

quinnj (approved):

Issue #27, opened 2020-10-27
Deprecate `All`


pdeffebach:
Maybe I'm missing something but there seems to be support in 

* DataFrames: https://github.com/JuliaData/DataFrames.jl/issues/2203
* DataAPI: https://github.com/JuliaData/DataAPI.jl/issues/16

for deprecating `All` in favor of `Cols`. 

* `Cols()`: No columns
* `Cols(:)` All columns

This is so easy with `Cols` that there is no need for `All()`. 

We should deprecate `All`.

bkamins:
Actually `Cols(:)` is not needed as it is just `:`.

@nalimilan - are you OK to deprecate `All`. I am OK with this.

bkamins:
The issue is that JuliaDB.jl would have to sync with this. CC @piever

nalimilan:
Yes, this is the last chance to deprecate it.

piever:
For me to understand, the key idea is that `Cols(args...)` can take a list of selectors and returns a selector corresponding to the union of the selections. So for JuliaDB to support this, `Cols` should be implemented the same as `All`, except for the empty case.

On top of that, `select(t, :)` should select all columns, that is, `:` is itself a "special selector". Supporting arbitrary ranges, like `select(t, 1:3)`, would be a bit messy, as they currently have a different meaning, that is `select(t, v::AbstractVector) == v`). OTOH, just `select(t, :)` seems reasonable.

In terms of what's breaking though, I would argue that adding this depwarn to DataAPI is breaking, as it would cause a nonsensical warning on older versions of packages that implement `All` but not yet `Cols`.  Maybe it'd be best to bump DataAPI version when introducing the depwarn, but not when removing `All`.

bkamins:
>  Maybe it'd be best to bump DataAPI version when introducing the depwarn

This is the plan.

> So for JuliaDB to support this, Cols should be implemented the same as All, except for the empty case.

Yes

> `:`

You do not have to support it, but it feels natural to allow `:`, as it is a common pattern in Julia Base and other languages. `Colon()` is is not `AbstractVector`.

piever:
Well, if we do not support it, then `Cols(:)` (which is the recommended deprecation) would not work out of the box, so that would introduce other special casing / inconsistencies. I actually wondered why the depwarn does not just recommend `:`.

[...]

PR #28, opened 2020-10-27
Deprecate All

bkamins:


bkamins:
Fixes https://github.com/JuliaData/DataAPI.jl/issues/27

bkamins:
I left the tests for `All` for the time being as is (I propose to remove them when we remove `All`).

src/DataAPI.jl
```diff
@@ -121,7 +121,14 @@ Select the union of the selections in `cols`. If `cols == ()`, select all column
 """
 struct All{T<:Tuple}
     cols::T
-    All(args...) = new{typeof(args)}(args)
+    function All(args...)
+        if isempty(args)
+            Base.depwarn("All() is deprecated, use Cols(:) instead", :All)
+        else
```
bkamins, line 127:
```suggestion
        if !isempty(args)
```

bkamins:
per recommendation of @piever I updated the PR to allow `All()` as I think it is not a problem to keep it.

nalimilan (approved):
Looks good, though better wait a bit to see whether IndexedTable maintainers are OK with deprecating `All()` too.

bkamins:
It seems we will not get much feedback here. Given `All()` is a natural name I would recommend to retain `All()` but deprecate `All(args...)`.

piever (approved):

bkamins:
@nalimilan - do we go ahead and merge this then?

bkamins:
Thank you. Will you tag a release?

PR #31, opened 2020-11-22
Add droplevels! function for removing unused pool values

quinnj:


src/DataAPI.jl
```diff
@@ -69,6 +69,20 @@ default definition.
 function refpool end
 refpool(A::AbstractArray) = nothing
 
+"""
+    droplevels!(A)
+
+For an array `A` that implements [`levels`](@ref DataAPI.levels)), certain levels may
+become "unused" via array modification, meaning the values still exist in the array
+"pool", but are no longer present in the actual array values. `droplevels!` remove
```
bkamins, line 77: maybe add a reference to `refpool` explicitly rather than `"pool"`?

src/DataAPI.jl
```diff
@@ -69,6 +69,20 @@ default definition.
...
+such values to ensure `levels` only returns values that currently exist in the array.
+
+This generic function is owned by DataAPI.jl itself, which is the sole provider of the
```
bkamins, line 80: in CategoricalArrays.jl `droplevels!` returns `A`. @nalimilan probably can comment better here about the design assumptions.

PR #32, opened 2021-01-24
Fix levels on Julia 1.7

nalimilan:
`unique` now returns a range for range inputs, so we cannot update it in place.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -95,7 +95,7 @@ function levels(x)
     u = unique(x)
     # unique returns its input with copying for ranges
     # (and possibly for other types guaranteed to hold unique values)
-    nmu = u === x ? filter(!ismissing, u) : filter!(!ismissing, u)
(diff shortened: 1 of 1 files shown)


src/DataAPI.jl
```diff
@@ -92,7 +92,11 @@ actually occur in the data, and does not preserve their order of appearance in `
 """
 function levels(x)
     T = Base.nonmissingtype(eltype(x))
-    levs = convert(AbstractArray{T}, filter!(!ismissing, unique(x)))
+    u = unique(x)
+    # unique returns its input with copying for ranges
+    # (and possibly for other types guaranteed to hold unique values)
+    nmu = u === x ? filter(!ismissing, u) : filter!(!ismissing, u)
```
bkamins, line 98: I am not sure if this is needed, but it would add an additional level of safety (e.g. if `unique` returned a `view`)
```suggestion
    nmu = (u === x || Base.mightalias(u, x)) ? filter(!ismissing, u) : filter!(!ismissing, u)
```

bkamins (approved):

quinnj (approved):

PR #33, opened 2021-01-29
Add invrefpool

bkamins:
Having a generic access to `invrefpool` is needed in https://github.com/JuliaData/DataFrames.jl/pull/2612.

Consider a short table and a long table joined on some column. In order to be fast we need to map values from short table key to ref values of long table key. This allows two things for `innerjoin`:
1. we immediately can drop values from short table not present in long table.
2. later we can do join on integer columns which is way faster than joining on e.g. string column.

Also since we do mapping of short table this operation should be fast.

In particular if short table defines `refarray` it is particularly fast, as we only need to map the reference values.

For CategoricalArrays.jl and PooledArrays.jl `invrefpool` is simply `get` on the inverted pool `Dict` with `nothing` as a sentinel.

I am not sure what would have to be defined in Arrow.jl.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -70,6 +70,23 @@ function refpool end
 refpool(A::AbstractArray) = nothing
 
 """
+    invrefpool(A)
+
+Whenever available, return an indexable object `invpool` such that, given the *original*
+array `A` and a "value" `x`, `pool(A)[invpool(A)[x]]` is equal to `x`.
+Return `nothing` if such "ref value" is not available.
+
+By default, `refpool(A)` returns `nothing`.
+
+If `invrefpool(A)` is not `nothing`, then `pool(A)` also muts not be `nothing`.
+
+This generic function is owned by DataAPI.jl itself, which is the sole provider of the
+default definition.
+"""
+function invrefpool end
+invrefpool(A::AbstractArray) = nothing
+
+"""
     describe(io::IO, x)
 
 For an object `x`, print descriptive statistics to `io`.
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -74,7 +74,7 @@ refpool(A::AbstractArray) = nothing
 
 Whenever available, return an indexable object `invpool` such that, given the *original*
 array `A` and a "value" `x`, `pool(A)[invpool(A)[x]]` is equal to `x`.
-Return `nothing` if such "ref value" is not available.
+`invpool(A)[x]` returns `nothing` if such `x` is not available.
 
 By default, `refpool(A)` returns `nothing`.
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -72,12 +72,16 @@ refpool(A::AbstractArray) = nothing
 """
     invrefpool(A)
 
-Whenever available, return an indexable object such that given an array `A` for which `refpool(A)` is not `nothing` such that:
+Whenever available, return an indexable object such that given an array `A`
+for which `refpool(A)` is not `nothing`:
 
-* for any valid index `x` into `refpool(A)` the following holds `invrefpool(A)[refpool(A)[x]] === x`;
-* for any valid index `ix` into `invrefpool(A)` the following holds `refpool(A)[invrefpool(A)[ix]] === ix`.
+* for any valid index `x` into `refpool(A)`, `invrefpool(A)[refpool(A)[x]]` is equal to `x`
+  (according to `isequal`) and of the same type as `x`;
+* for any valid index `ix` into `invrefpool(A)` , `refpool(A)[invrefpool(A)[ix]]` is equal to `ix`
+  (according to `isequal`) and of the same type as `ix`.
 
-Additionally it is requred that `haskey` is defined for `invrefpool(A)` allowing to check if `ix` is a valid index into it.
+Additionally it is required that `haskey` is defined for `invrefpool(A)`,
+allowing to check if `ix` is a valid index into it.
 
 By default, `refpool(A)` returns `nothing`.
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -72,13 +72,14 @@ refpool(A::AbstractArray) = nothing
 """
     invrefpool(A)
 
-Whenever available, return an indexable object `invpool` such that, given the *original*
-array `A` and a "value" `x`, `pool(A)[invpool(A)[x]]` is equal to `x`.
-`invpool(A)[x]` returns `nothing` if such `x` is not available.
+Whenever available, return an indexable object such that, given an array `A` and
+a "value" `x` taken from `A`, `refpool(A)[invrefpool(A)[x]]` is equal to `x`
+(according to `isequal`) and of the same type as `x`.
+`invrefpool(A)[x]` returns `nothing` if such `x` is not available.
 
 By default, `refpool(A)` returns `nothing`.
 
-If `invrefpool(A)` is not `nothing`, then `pool(A)` also muts not be `nothing`.
+If `invrefpool(A)` is not `nothing`, then `pool(A)` also must not be `nothing`.
 
 This generic function is owned by DataAPI.jl itself, which is the sole provider of the
 default definition.
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -83,7 +83,7 @@ for which `refpool(A)` is not `nothing`:
 Additionally it is required that `haskey` is defined for `invrefpool(A)`,
 allowing to check if `ix` is a valid index into it.
 
-By default, `refpool(A)` returns `nothing`.
+By default, `invrefpool(A)` returns `nothing`.
 
 If `invrefpool(A)` is not `nothing`, then `refpool(A)` also must not be `nothing`.
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -72,14 +72,16 @@ refpool(A::AbstractArray) = nothing
 """
     invrefpool(A)
 
-Whenever available, return an indexable object such that, given an array `A` and
-a "value" `x` taken from `A`, `refpool(A)[invrefpool(A)[x]]` is equal to `x`
-(according to `isequal`) and of the same type as `x`.
-`invrefpool(A)[x]` returns `nothing` if such `x` is not available.
+Whenever available, return an indexable object such that given an array `A` for which `refpool(A)` is not `nothing` such that:
+
+* for any valid index `x` into `refpool(A)` the following holds `invrefpool(A)[refpool(A)[x]] === x`;
+* for any valid index `ix` into `invrefpool(A)` the following holds `refpool(A)[invrefpool(A)[ix]] === ix`.
+
+Additionally it is requred that `haskey` is defined for `invrefpool(A)` allowing to check if `ix` is a valid index into it.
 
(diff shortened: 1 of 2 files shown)


bkamins:
Also could we add a restriction that ref values have to be non-negative (I am not sure if for Arrow.jl it would be acceptable). This would simplify code for me as negative value could be used as a sentinel (just like in `Dict` in Base, when negative value is internally returned when lookup fails).

src/DataAPI.jl
```diff
@@ -69,6 +69,23 @@ default definition.
 function refpool end
 refpool(A::AbstractArray) = nothing
 
+"""
+    invrefpool(A)
+
+Whenever available, return an indexable object `invpool` such that, given the *original*
+array `A` and a "value" `x`, `pool(A)[invpool(A)[x]]` is equal to `x`.
+Return `nothing` if such "ref value" is not available.
```
bkamins, line 77:
```suggestion
`invpool(A)[x]` returns `nothing` if such `x` is not available.
```

bkamins:
Ah - now I have realized that we even do not require "ref value" to be an integer. So this is a question again if we want to add such a restriction. If not then `invrefpool(A)` should support not only `getindex` but also `haskey` and `get`.

At least I would add a restriction that some sentinel, e.g. `nothing` is not allowed to be "ref value".

What do you think?

src/DataAPI.jl
```diff
@@ -69,6 +69,23 @@ default definition.
...
+"""
+    invrefpool(A)
+
+Whenever available, return an indexable object `invpool` such that, given the *original*
+array `A` and a "value" `x`, `pool(A)[invpool(A)[x]]` is equal to `x`.
+`invpool(A)[x]` returns `nothing` if such `x` is not available.
+
+By default, `refpool(A)` returns `nothing`.
+
+If `invrefpool(A)` is not `nothing`, then `pool(A)` also muts not be `nothing`.
```
nalimilan, line 81:
```suggestion
If `invrefpool(A)` is not `nothing`, then `pool(A)` also must not be `nothing`.
```

src/DataAPI.jl
```diff
@@ -69,6 +69,23 @@ default definition.
 function refpool end
 refpool(A::AbstractArray) = nothing
 
+"""
+    invrefpool(A)
+
+Whenever available, return an indexable object `invpool` such that, given the *original*
+array `A` and a "value" `x`, `pool(A)[invpool(A)[x]]` is equal to `x`.
+`invpool(A)[x]` returns `nothing` if such `x` is not available.
```
nalimilan, line 77:
```suggestion
Whenever available, return an indexable object such that, given an array `A` and
a "value" `x` taken from `A`, `refpool(A)[invrefpool(A)[x]]` is equal to `x`
(according to `isequal`) and of the same type as `x`.
`invrefpool(A)[x]` returns `nothing` if such `x` is not available.

**Review by nalimilan — COMMENTED** [review]
Makes sense. Two points:
- Do we really need to require `invrefpool(A)[x]` to return `nothing` when `x` isn't found?
- Maybe we should actually require that `invrefpool(A)[A[i]]` give the same value as `refarray(A)[i]`? That would be stricter, and is probably what we expect. Maybe you even rely on that in your PR?

**Comment by bkamins** [comment]
> Do we really need to require `invrefpool(A)[x]` to return nothing when x isn't found?

No. We just need some way to tell if `x` is in the pool or not. Therefore requiring that `haskey` or supporting `get` is an alternative.

**Comment by nalimilan** [comment]
> No. We just need some way to tell if `x` is in the pool or not. Therefore requiring that `haskey` or supporting `get` is an alternative.

OK, then maybe requiring `haskey` is better. Otherwise that prevents implementing `invrefpool` using a standard `Dict`.

What do you think about my other point?


> Also could we add a restriction that ref values have to be non-negative (I am not sure if for Arrow.jl it would be acceptable). This would simplify code for me as negative value could be used as a sentinel (just like in `Dict` in Base, when negative value is internally returned when lookup fails).
>
> Ah - now I have realized that we even do not require "ref value" to be an integer. So this is a question again if we want to add such a restriction. If not then `invrefpool(A)` should support not only `getindex` but also `haskey` and `get`.
> 
> At least I would add a restriction that some sentinel, e.g. `nothing` is not allowed to be "ref value".
> 
> What do you think?

Sorry I had missed these comments. This is an orthogonal issue so we can maybe discuss it separately. Maybe you could use `first(refpool(A))-1` as a sentinel?

File: src/DataAPI.jl [hunk 4]
```diff
@@ -69,6 +69,24 @@ default definition.
 function refpool end
 refpool(A::AbstractArray) = nothing
 
+"""
+    invrefpool(A)
+
+Whenever available, return an indexable object such that, given an array `A` and
+a "value" `x` taken from `A`, `refpool(A)[invrefpool(A)[x]]` is equal to `x`
+(according to `isequal`) and of the same type as `x`.
+`invrefpool(A)[x]` returns `nothing` if such `x` is not available.
```
bkamins, line 78:
```suggestion
Whenever available, return an indexable object such that given an array `A` for which `refpool(A)` is not `nothing` such that:
* for any valid index `x` into `refpool(A)` the following holds `invrefpool(A)[refpool(A)[x]] === x`;
* for any valid index `ix` into `invrefpool(A)` the following holds `refpool(A)[invrefpool(A)[ix]] === ix`.
```

src/DataAPI.jl
```diff
@@ -69,6 +69,24 @@ default definition.
...
+
+By default, `refpool(A)` returns `nothing`.
+
+If `invrefpool(A)` is not `nothing`, then `pool(A)` also must not be `nothing`.
```
bkamins, line 82:
```suggestion
If `invrefpool(A)` is not `nothing`, then `refpool(A)` also must not be `nothing`.
```

bkamins:
I would go for `haskey` then. We cannot use `first(refpool(A))-1` for sentinel as:
1. we do not have a guarantee that references are numbers now.
2. even if we had this guarantee subtracting `1` could be problematic in theory (e.g. for unsigned values we could get `-1` and type instability)

But I worked around it and `haskey` is enough for me.

As for:

> `invrefpool(A)[A[i]]` give the same value as `refarray(A)[i]`

we can add also this requirement, but it is actually looser than what I required I think (as `refpool` can contain values not present in `A` and we still want `invrefpool` should also have a mapping for it).

I have updated the docstring to be precise. Hopefully you are OK with it.

src/DataAPI.jl
```diff
@@ -69,6 +69,26 @@ default definition.
...
 
+"""
+    invrefpool(A)
+
+Whenever available, return an indexable object such that given an array `A` for which `refpool(A)` is not `nothing` such that:
+
+* for any valid index `x` into `refpool(A)` the following holds `invrefpool(A)[refpool(A)[x]] === x`;
+* for any valid index `ix` into `invrefpool(A)` the following holds `refpool(A)[invrefpool(A)[ix]] === ix`.
+
+Additionally it is requred that `haskey` is defined for `invrefpool(A)` allowing to check if `ix` is a valid index into it.
```
nalimilan, line 80: Currently we don't require `=== ` in the docstring for `refpool` (only `isequal` and equal type). We could discuss changing this if we have a reason to, but I'm not sure we have.
```suggestion
Whenever available, return an indexable object such that given an array `A`
for which `refpool(A)` is not `nothing`:

* for any valid index `x` into `refpool(A)`, `invrefpool(A)[refpool(A)[x]]` is equal to `x`
  (according to `isequal`) and of the same type as `x`;
* for any valid index `ix` into `invrefpool(A)` , `refpool(A)[invrefpool(A)[ix]]` is equal to `ix`
  (according to `isequal`) and of the same type as `ix`.

Additionally it is required that `haskey` is defined for `invrefpool(A)`,
allowing to check if `ix` is a valid index into it.
```

nalimilan:
> I would go for `haskey` then. We cannot use `first(refpool(A))-1` for sentinel as:
> 
>     1. we do not have a guarantee that references are numbers now.
> 
>     2. even if we had this guarantee subtracting `1` could be problematic in theory (e.g. for unsigned values we could get `-1` and type instability)
> 
> 
> But I worked around it and `haskey` is enough for me.

OK. Actually what I do in the DataFrames grouping code is that I only use the fast path when `refarray(A) isa AbstractArray{Integer}`. That way we're safe but we still allow other types in case that could be useful for some packages.

bkamins:
> That way we're safe but we still allow other types in case that could be useful for some packages.

I agree. We just need to remember about it.

> Currently we don't require `===` in the docstring for refpool (only `isequal` and equal type)

I know, but what is the benefit of not requiring `===`? (we can discuss it in other PR, I accepted your suggestion in this PR).

If `isequal` is implemented correctly "`isequal` and equal type" should be equivalent to `===` for immutable types.
And I think we should disallow mutable types in PooledArrays.jl (we already disallow them in CategoricalArrays.jl). I am not sure about Arrow.jl (cc @quinnj). See https://github.com/JuliaData/PooledArrays.jl/issues/51.

nalimilan:
> I know, but what is the benefit of not requiring `===`? (we can discuss it in other PR, I accepted your suggestion in this PR).
> 
> If `isequal` is implemented correctly "`isequal` and equal type" should be equivalent to `===` for immutable types.
> And I think we should disallow mutable types in PooledArrays.jl (we already disallow them in CategoricalArrays.jl). I am not sure about Arrow.jl (cc @quinnj). See [JuliaData/PooledArrays.jl#51](https://github.com/JuliaData/PooledArrays.jl/issues/51).

I'm not sure there's a benefit, but I'm not sure there's a benefit to requiring `===` either. :-) Anyway it's not equivalent, for example:
```julia
julia> NaN === 0/0
false

julia> isequal(NaN, 0/0)
true
```
More generally, any immutable type can implement `isequal` in a less strict way than `===`, e.g. if it includes references to mutable objects like `CategoricalValue`. There may also be corner cases with strings.

src/DataAPI.jl
```diff
@@ -69,6 +69,30 @@ default definition.
...
+
+* for any valid index `x` into `refpool(A)`, `invrefpool(A)[refpool(A)[x]]` is equal to `x`
+  (according to `isequal`) and of the same type as `x`;
+* for any valid index `ix` into `invrefpool(A)` , `refpool(A)[invrefpool(A)[ix]]` is equal to `ix`
+  (according to `isequal`) and of the same type as `ix`.
+
+Additionally it is required that `haskey` is defined for `invrefpool(A)`,
+allowing to check if `ix` is a valid index into it.
+
+By default, `refpool(A)` returns `nothing`.
```
nalimilan, line 86:
```suggestion
By default, `invrefpool(A)` returns `nothing`.
```

nalimilan (approved):

bkamins:
Ah yes - the `NaN` thing is nasty.

@quinnj - is it OK to merge it as is? The next step would be to add the support in PooledArrays.jl, CategoricalArrays.jl and Arrow.jl and make their releases.

quinnj (approved):
Yes, LGTM.

I've opened https://github.com/JuliaData/Arrow.jl/issues/120 since we haven't implemented any DataAPI methods for `Arrow.DictEncoded` yet and I keep forgetting about it.

bkamins:
OK - I will wait till tomorrow and if there are no more comments then merge this PR and tag a release. Then I would open PRs to PooledArrays.jl and CategoricalArrays.jl to provide the API.

PR #35, opened 2021-02-14
Add DataAPI.unwrap with default definition

quinnj:
As discussed in
https://github.com/JuliaData/CategoricalArrays.jl/issues/142.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -85,4 +85,9 @@ end
 
 end
 
+@testset "unwrap" begin
+    @test DataAPI.unwrap(1) === 1
+    @test DataAPI.unwrap(missing) === missing
+end
+
 end # @testset "DataAPI"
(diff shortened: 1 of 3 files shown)


bkamins:
Thank you. Could you also bump a version of DataAPI.jl so that we could register it after merging? Thank you.

bkamins (approved):

test/runtests.jl
```diff
@@ -85,4 +85,9 @@ end
 
 end
 
+@testset "unwrap" begin
+    @test DataAPI.unwrap(1) === 1
+    @test DataAPI.unwrap(missing) === missing
```
nalimilan, line 90: Ah, clever: with that fallback we don't need to use `passmissing(unwrap)`!

nalimilan (approved):
Thanks!

test/runtests.jl
bkamins: Yes!

PR #36, opened 2021-05-02
Add `join` family without default definition

iblis17:
close https://github.com/JuliaData/Tables.jl/issues/238

Well, I think the default definition isn't trivial to implement, so I just setup the interface functions.
I don't restrict the parameter of join key, which is the kwarg `on` in DataFrames.jl, since the join key is optional in some cases of time series join.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -183,128 +183,16 @@ default definition.
 function unwrap end
 unwrap(x) = x
 
-const _joindoc = """
-The implementation specific arguments, like the join key or equality rule,
-can be designed as the additional keyword arguments.
-"""
-
-"""
-    innerjoin(x, y)
-
-Perform a database-style inner join of two tabular objects and return a tabular
-object containing the result.
-
-$_joindoc
-
-See also: [`outerjoin`](@ref),
-          [`leftjoin`](@ref),
-          [`rightjoin`](@ref),
-          [`semijoin`](@ref),
-          [`antijoin`](@ref),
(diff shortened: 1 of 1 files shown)


src/DataAPI.jl
```diff
@@ -183,4 +183,102 @@ default definition.
 function unwrap end
 unwrap(x) = x
 
+"""
+    innerjoin(x, y, zs...)
```
bkamins, line 187: Two small comments:
1. Here and below I would drop `zs`, as other packages do not have to provide this.
2. I think it would be good to mention a reference (or add a description) to what a given type of join does. Ideally (if it is possible to do it relatively easily and in a generic way) - maybe also add a comment that the way key columns are determined is implementation specific and that key equality rules are also implementation specific.

nalimilan: Actually I'm not sure it's very useful to provide docstrings. Packages will have to provide more specific docstrings anyway, and the default docstrings are even misleading as `on` may be required by some packages.

For `pairwise` in StatsAPI I just added a comment for the same reason. I think that makes sense since that documentation is intended at package authors rather than end users.

bkamins: I am OK to fully remove docstrings.

bkamins (approved):

src/DataAPI.jl
iblis17: I remove the docstrings and add some code comments, since it's face to developers.

nalimilan (approved):

PR #40, opened 2021-07-22
add nrow and ncol

bkamins:
Following the discussion in https://github.com/JuliaAI/ScientificTypes.jl/issues/137#issuecomment-885258513.

CC @quinnj @nalimilan @ablaom

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DataAPI"
 uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.7.0"
+version = "1.8.0"
 
 [compat]
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DataAPI"
 uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.8.0"
+version = "1.9.0"
 
 [compat]
 julia = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DataAPI"
 uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.9.0"
+version = "1.8.0"
 
 [compat]
 julia = "1"
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -205,7 +205,7 @@ function nrow end
 """
     ncol(t)
 
-Return the number of rows of table `t`.
+Return the number of columns of table `t`.
 """
 function ncol end
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -195,4 +195,18 @@ function semijoin end
 function antijoin end
 function crossjoin end
 
+"""
+    nrow(t)
+
+Return the number of rows of table `t`.
+"""
+function nrow end
+
+"""
+    ncol(t)
+
+Return the number of rows of table `t`.
+"""
+function ncol end
+
 end # module


src/DataAPI.jl
```diff
@@ -195,4 +195,18 @@ function semijoin end
...
+    nrow(t)
+
+Return the number of rows of table `t`.
+"""
+function nrow end
+
+"""
+    ncol(t)
+
+Return the number of rows of table `t`.
```
bkamins, line 208:
```suggestion
Return the number of columns of table `t`.
```

nalimilan:
Makes sense. Maybe we should also say that Tables.jl is the owner of the function, and add a fallback definition there (that would notably work for vectors of names tuples)? I'm also not sure whether we should provide a docstring here, or in Tables.jl instead, given that we don't define any method here (cf. discussion regarding join functions).

bkamins:
@quinnj - is making Tables.jl an owner feasible. I was thinking about it also, but I was not sure.

An alternative would be to make the default implementation in DataAPI.jl that would return `missing` (signaling that it is unknown)?
Then clearly Tables.jl could add extra methods in cases when `nrow` or `ncol` is known.

nalimilan:
I'd rather throw a `MethodError` than return `missing` for unsupported types. The latter could be useful for table objects for which the number of rows cannot be known in advance (e.g. iterators of named tuples).

bkamins:
OK - makes sense. With the definitions I gave now `MethodError` is thrown appropriately.
So I assume that if we merge and release it Tables.jl should add `nrow` and `ncol` definitions to the tables it ships. Right?

bkamins:
Ah - and I made the docsting, as as opposed to `join*` case I think we have a fixed API here (as opposed to `join*` where different packages can probably have different args/kwargs)

ablaom:
Just FYI: For a Tables.jl implementation one might look at what we have currently in MLJBase.jl (courtesy of @OkonSamuel): https://github.com/JuliaAI/MLJBase.jl/blob/f21d295d3f0714ec8f5dcc93d1aa34ccbf16fdff/src/interface/data_utils.jl#L77 .

quinnj:
I'm fine with adding these definitions here; I've mentioned elsewhere in various Tables.jl issues, however, that `nrow` and `ncol` aren't well-defined for all "tables" (in the Tables.jl source definition). You can have a row-iterator table with `Base.SizeUnknown()` or a filtering generator of NamedTuples and you won't be able to call `nrow` on those.

Now, I will say that's just the row-based table case. For column-based tables, we do expect the # of rows to be defined, and Tables.jl has `Tables.rowcount` internally defined that works on any result of `Tables.columns(x)`.

And so, I'd say I don't think I'm on board with having Tables.jl be the owner, because I'm afraid that will give the impression that `nrow` is now supported on any table. It makes it tricky though because to rename `Tables.rowcount` to `nrow`, it kind of needs to be the fallback definition. It's a little messy.

bkamins:
So - in summary. What I propose now is just what is needed if I understand the comments correctly. Let us wait for @nalimilan to review.

bkamins:
@nalimilan - could you please review. After merging this I would make a 1.9 release. Thank you!

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DataAPI"
 uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.7.0"
+version = "1.9.0"
```
nalimilan, line 4:
```suggestion
version = "1.8.0"
```

nalimilan (approved):

bkamins:
Thank you! Making a release

ablaom:
cc @OkonSamuel

matthieugomez:
Would not `nrows` and `ncols` be better names? It would be more consistent with Julia Base (e.g. `ndims`, `nfields`).

bkamins:
I think it is too late to change this for DataFrames.jl, and given we will stick to `nrow` and `ncol` there it is better to use a common naming scheme across packages.

PR #42, opened 2021-09-20
Extend Cols API

bkamins:
Following https://github.com/JuliaData/DataFrames.jl/pull/2881/files#r711943431.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -167,7 +167,6 @@ Select the union of the selections in `cols`. If `cols == ()`, select no columns
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
-As a special case if `:` is passed (the `Colon` function) select all columns.
 """
 struct Cols{T<:Tuple}
     cols::T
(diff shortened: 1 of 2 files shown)


src/DataAPI.jl
```diff
@@ -161,8 +161,13 @@ end
 
 """
     Cols(cols...)
+    Cols(f::Function)
 
 Select the union of the selections in `cols`. If `cols == ()`, select no columns.
+
+If the only positional argument is a `Function` `f` then select the columns whose
+names passed to the `f` predicate as strings return `true`.
+As a special case if `:` is passed (the `Colon` function) select all columns.
```
nalimilan, line 170: Do we really need to mention `:`? It sounds quite obvious that if `:` is allowed as a selector to get all columns, passing it selects all columns.
```suggestion
```

bkamins: Technically `:` is a `Function` so this is a special case. Do you still think it should be removed?

nalimilan: Ah, got it! I wouldn't mention it though, as nobody would expect `:` to be treated as a predicate on strings (anyway that would error).

bkamins: OK - removing.

nalimilan (approved):

bkamins:
Thank you!

Issue #44, opened 2022-03-11
add kwarg to levels to keep missing


bkamins:
See https://github.com/JuliaData/DataFrames.jl/pull/3012#discussion_r814268223

PR #46, opened 2022-03-20
Add `skipmissing` argument to `levels`

nalimilan:
Currently if one wants to get levels in their appropriate order, but add `missing` if present, the only solution is to do something like `union(levels(x), unique(x))`, which is inefficient.
Support `skipmissing=false` to allow doing this in a single pass over the data.

Use `@inline` to ensure that the return type can be inferred when the value of `skipmissing` is known statically. Also fix a type instability which existed for ranges.

Of course this new feature won't work for custom types which override this method (like `CategoricalArray`) until packages implement it. Unfortunately there's no way for packages which would like to rely on it (like DataFrames) to require an appropriate version.

There will also be a decision to make in CategoricalArrays as to whether `missing` should be returned only when present in the data (like the method defined here) or all the time as long as the eltype allows for it (like for other levels, which is more efficient).

Fixes https://github.com/JuliaData/DataAPI.jl/issues/44.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -121,35 +121,47 @@ actually occur in the data, and does not preserve their order of appearance in `
 @inline levels(x; skipmissing::Bool=true) =
     skipmissing ? _levels_skipmissing(x) : _levels_missing(x)
 
+# The `which` check is here for backward compatibility:
+# if a type implements a custom `levels` method but does not support
+# keyword arguments, `levels(x, skipmissing=true/false)` will dispatch
+# to the fallback methods here, and we take care of calling that custom method
 function _levels_skipmissing(x)
-    T = Base.nonmissingtype(eltype(x))
-    u = unique(x)
-    # unique returns its input with copying for ranges
-    # (and possibly for other types guaranteed to hold unique values)
-    nmu = (u isa AbstractRange || u === x || Base.mightalias(u, x)) ?
-        filter(!ismissing, u) : filter!(!ismissing, u)
-    levs = convert(AbstractArray{T}, nmu)
-    try
-        sort!(levs)
-    catch
+    if which(DataAPI.levels, Tuple{typeof(x)}) === which(DataAPI.levels, Tuple{Any})
+        T = Base.nonmissingtype(eltype(x))
+        u = unique(x)
+        # unique returns its input with copying for ranges
+        # (and possibly for other types guaranteed to hold unique values)
+        nmu = (u isa AbstractRange || u === x || Base.mightalias(u, x)) ?
+            filter(!ismissing, u) : filter!(!ismissing, u)
+        levs = convert(AbstractArray{T}, nmu)
+        try
+            sort!(levs)
+        catch
+        end
+        return levs
+    else
+        return levels(x)
     end
-    levs
 end
 
 function _levels_missing(x)
-    levs = convert(AbstractArray{eltype(x)}, unique(x))
-    # unique returns its input with copying for ranges
-    # (and possibly for other types guaranteed to hold unique values)
-    docopy = x isa AbstractRange || levs === x || Base.mightalias(levs, x)
-    try
-        if docopy
-            levs = sort(levs)
-        else
+    if which(DataAPI.levels, Tuple{typeof(x)}) === which(DataAPI.levels, Tuple{Any})
+        u = convert(AbstractArray{eltype(x)}, unique(x))
+        # unique returns its input with copying for ranges
+        # (and possibly for other types guaranteed to hold unique values)
+        levs = (x isa AbstractRange || u === x || Base.mightalias(u, x)) ?
+            Base.copymutable(u) : u
+        try
             sort!(levs)
+        catch
         end
-    catch
+        return levs
+    # This is a suboptimal fallback since it does a second pass over the data
+    elseif any(ismissing, x)
+        return [levels(x); missing]
+    else
+        return convert(AbstractArray{eltype(x)}, levels(x))
     end
-    levs
 end
 
 """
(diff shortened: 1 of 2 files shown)


bkamins:
CI on 1.0 fails.

bkamins (commented):
Implementation looks good. The only problem is the one that you have noted - we cannot rely on this behavior in library codes anyway.
Maybe we can have this for user's code, but then add something like `levelsmissing` that is internal and intended for library code to use (your current `_levels_missing` which is internal but not indented to be used).

nalimilan:
> Maybe we can have this for user's code, but then add something like `levelsmissing` that is internal and intended for library code to use (your current `_levels_missing` which is internal but not indented to be used).

So the advantage would be that we would be sure that it doesn't throw an error? If a type that uses a custom order of levels doesn't implement it, it would return them in their order of appearance though. Not sure whether it's better to throw an explicit error so that users can complain or give them suboptimal results...

bkamins:
Having thought of it again, actually your solution is OK. We only would need in `_levels_missing` and `_levels_skipmissing` check if for the passed type the source package defines `levels`. If it does then `levels` would be called.

The point is that we would get to these calls only if package defines `levels(x::SomeType)` but does not define `levels(x::SomeType; skipmissing::Bool=true)`. The reason is that if it would define new API then we would never reach the internal methods. On the other hand if it does not define the new signature, but defines the old one we know it is `levels` without missing.

nalimilan:
Ah, interesting. You mean that `_levels_missing` would do something like this?
```
if which(DataAPI.levels, Tuple{typeof(x)}) === which(DataAPI.levels, Tuple{Vector})
    ... # Current code
elseif any(ismissing, col)
    [_levels_skipmissing(col); missing]
else
    _levels_skipmissing(col)
end
```

bkamins:
In `_levels_missing` I mean the following:
```
if which(DataAPI.levels, (typeof(x),_) === which(DataAPI.levels, (Any,))
    ... # Current code
elseif any(ismissing, col)
    [levels(col); missing]
else
    levels(col)
end
```

The point is that if `which(DataAPI.levels, (typeof(x),_) === which(DataAPI.levels, (Any,))` is `false` then "old style" `levels` is defined for `typeof(x)`.

nalimilan:
Ah right. Though doing this would mean that DataFrames would have to call `_levels_missing` rather than `levels` (otherwise the method defined by CategoricalArrays would be called, throwing an error since it doesn't accept `skipmissing` yet). I'd rather either 1) do the `which` check in DataFrames instead, or 2) bite the bullet and call the function `levels_missing` and document it.

bkamins:
No it would not.

Note what would happen under my approach if you called `levels(x, skipmissing=false)` in DataFrames.jl and `x` were `CategoricalArray` and it would not define `levels` with `skipmissing` option. The process would be the following:
1. Default `levels` would be called from DataAPI.jl (as `levels` from CategoricalArrays.jl does not define kwargs - this is one of these cases when kwargs influence dispatch)
2. then this default method would call `_levels_missing`
3. inside `_levels_missing` we would see that `levels` is defined for `CategoricalArray`
4. so `_levels_missing` would execute `[levels(x); missing]` if indeed there would be `missing` in `x`

This would not be super fast, but this would be a default fallback and I think it would be good enough.

However, your option 2 is I think also OK and we could just add a second function not to complicate things if you prefer so.

Having said that the question is if `CategoricalArray` allows `Missing` should `missing` be included in levels (as it is a potential level) or not? I see in your code that you decided that not, but this is not consistent with a general notion of `levels` as a function listing potentially possible values.

nalimilan:
OK, great, I hadn't realized that dispatch would choose the method which supports keyword arguments. I've pushed a commit to do that.

Unfortunately, `@inferred` seems to be confused by the recursive call to `levels`, as it thinks the result is type-unstable even though `@code_warntype` doesn't. I couldn't find a solution for now. Also the result is quite complex so I wonder whether it's worth it (DataAPI is supposed to be a lightweight package...). At any rate I'd have to add a test to ensure the backward compatibility works.

> Having said that the question is if `CategoricalArray` allows `Missing` should `missing` be included in levels (as it is a potential level) or not? I see in your code that you decided that not, but this is not consistent with a general notion of `levels` as a function listing potentially possible values.

Yes that's the debate I mentioned above. For now I didn't decide anything, given that we only define the fallback method here, which is documented to be equivalent to `unique` but with possibly a different order. CategoricalArrays can do something slightly different if appropriate. Though it's difficult to decide, as `missing` isn't really a level. I'm worried that including `missing` even when it's not present in the data would be annoying and unexpected. For example, if you call `completecases` to skip missing values, the eltype won't change, yet adding `missing` would probably not be what you want.

bkamins:
> Yes that's the debate I mentioned above

I know. I just did not want to forget about it. Let us agree that `missing` is included only if it is present and explicitly document this.

> Also the result is quite complex so I wonder whether it's worth it

I agree this is tricky (I just wanted to show it so that we consider such approach and decide if we want it). If you prefer - as I have written above - option 2, i.e. having two separate methods is I think also OK I think.

nalimilan:
OK, I've found a solution to make the function inferrable using a different type for the default value.

src/DataAPI.jl
```diff
@@ -105,29 +105,67 @@ definition.
...
+        try
+            sort!(levs)
+        catch
+        end
+        return levs
+    # This is a suboptimal fallback since it does a second pass over the data
+    elseif any(ismissing, x)
+        return [levels(x); missing]
+    else
+        return convert(AbstractArray{eltype(x)}, levels(x))
```
bkamins, line 167: this line seems not to be covered by tests

bkamins (approved):
Looks good. Thank you. Only test coverage needs improvement.

quinnj (approved):
LGTM; if you get a chance, would you mind running Arrow.jl test suite to ensure everything works there?

nalimilan:
Test pass.

PR #47, opened 2022-04-21
add allcombinations

bkamins:
add `allcombinations` function. see https://github.com/JuliaData/DataFrames.jl/pull/3031 for an example implementation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DataAPI"
 uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
 authors = ["quinnj <quinn.jacobd@gmail.com>"]
-version = "1.9.0"
+version = "1.10.0"
 
 [compat]
 julia = "1"
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -279,4 +279,12 @@ Return the number of columns of table `t`.
 """
 function ncol end
 
+"""
+    allcombinations(sink, ...)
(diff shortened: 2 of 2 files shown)


bkamins:
@nalimilan - now I realized that the approach proposed here (and discussed on slack) cannot be used.
The reason is that we cannot dispatch on `DataFrame` since its type is `DataType` (it is not a function).

Maybe we should leave https://github.com/JuliaData/DataFrames.jl/pull/3031 as it is then? Or do you see any other solution? The only I can see is that `allcombinations` would define a table-like type that then would be passed to `DataFrame` constructor. But then we would need to provide implementation of the logic of the function in DataAPI.jl, which I think we should not do.

nalimilan:
Can't we just define methods like `allcombinations(::Type{DataFrame}, ...)`?

bkamins:
Ah - right. We can 😄. In this case can you please review this PR?

nalimilan (approved):

nalimilan:
@quinnj We could mention in the docstring that TableOperations owns `allcombinations(NamedTuple, ...)` and implement a method there.

bkamins:
@quinnj - could you have a look at this PR before I merge it and make a release (and finish https://github.com/JuliaData/DataFrames.jl/pull/3031 a follow-up PR in DataFrames.jl)? Thank you!

quinnj (approved):

bkamins:
Thank you!

Issue #22, opened 2020-05-23
`metadata` method


Tokazama:
It is often the case that one wants to attach metadata of some sort to an array/graph/etc. How do people feel about adding something basic like `metadata(x) = nothing` that can then be extended by other packages?

bkamins:
@pdeffebach had some good design ideas about it in DataFrames.jl in the past.
Now, finally after 0.21.0 release, we are planning to add this functionality to DataFrames.jl.

As this is raised on a higher level let me give the API I envision for DataFrames.jl for now:
* `metadata(::DataFrame)` returns a `Union{Nothing, Dict{Symbol,Any}}` that - if filled - gives a `DataFrame`-level metadata (this can be arbitrary metadata). The restriction would be that symbols starting with `DF_` in their name would be reserved for internal use of DataFrames.jl (as a convention)
* `metadata(obj::OTHER_TYPES_DEFINED_BY_DATAFRAMES) = metadata(parent(obj))`
* `metadata(::DataFrame, ::ColumnIndex)` returns a `String` (by default nothing) - which would indicate just a verbose name of the column, with default being just column name
* `metadata(obj::OTHER_TYPES_DEFINED_BY_DATAFRAMES, ::ColumnIndex)` similar to the above if the column is present in the other type.

If we agree to this design then I can implement it. The key challenge is rules of propagation of metadata, but this is not DataAPI.jl related thing so I leave this discussion for later.

CC @pdeffebach, @nalimilan

nalimilan:
See https://github.com/JuliaData/DataFrames.jl/pull/1458 for the last attempt at implementing this in DataFrames. Two points:
- I think we need something more general than just having a custom label/verbose name for columns. For example it could be useful to store units, informations about measurement, etc. Label can just be a standard field among others.
- We also need an API to set metadata and to retrieve the list of fields that have been set.

In general a choice has to be made between having 1) a single function in the API which would return a metadata dict which would have to implement specific methods (`getindex`, `setindex!` and `keys` notably); or 2) several functions in the API that would allow doing these operations directly. See the table in my first comment at https://github.com/JuliaData/DataFrames.jl/pull/1458. I think returning an object is simpler since it allows reusing the standard dict API.

Tokazama:
[...]

PR #48, opened 2022-05-08
add `metadata`

bkamins:
Following the discussion in https://github.com/JuliaData/DataFrames.jl/issues/2961 I propose to have `getmetadata` be a function defined on DataAPI.jl level. In this way in particular:
* Arrow.jl take any table and if it supports `getmetadata` write appropriate metadata to arrow file;
* DataFrames.jl `DataFrame` constructor taking a table can check if it supports metadata and if it does automatically attach this metadata to a `DataFrame`;

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -331,12 +331,22 @@ Return `true` if column `column` of Tables.jl table `table` has non-empty metada
 and return `false` if metadata is empty.
 Return `nothing` if metadata is not supported for `column`.
 
-
 `column` should be any accepted column identifier by `table`, but at least `Symbol` must be supported.
 
 If `hascolmetadata` returns `Bool` then it is guaranteed that call to
 `colmetatada(table, columns)` will not throw an error.
 """
 hascolmetadata(::Any, ::Any) = nothing
+
+"""
+    hascolmetadata(table)
+
+Return `true` if there is exists at least one column `column` of Tables.jl table
+`table` for which `hascolmetadata(table, column)` returns `true`; otherwise
+return `false`.
+
+Return `nothing` if metadata is not supported.
+"""
+hascolmetadata(::Any) = nothing
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -338,6 +338,19 @@ metadata!(::T, ::AbstractString, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
 
 """
+    deletemetadata!(x, [key::AbstractString])
+
+Delete metadata for object `x` for key `key` and return `x`
+(if metadata for `key` is not present do not perform any action).
+If `key` is not passed delete all metadata for object `x`.
+If `x` does not support metadata deletion throw `ArgumentError`.
+"""
+deletemetadata!(::T, ::AbstractString) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
+deletemetadata!(::T) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
+
+"""
     colmetadata(x, col, key::AbstractString; style::Bool=false)
 
 Return metadata value associated with table `x` for column `col` and key `key`.
@@ -383,7 +396,7 @@ colmetadatakeys(::Any) = ()
     colmetadata!(x, col, key::AbstractString, value; style)
 
 Set metadata for table `x` for column `col` for key `key` to have value `value`
-and style `style`.
+and style `style` and return `x`.
 If `x` does not support setting metadata for column `col` throw `ArgumentError`.
 
 $COL_INFO
@@ -394,5 +407,25 @@ colmetadata!(::T, ::Int, ::AbstractString, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
 colmetadata!(::T, ::Symbol, ::AbstractString, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
+
+"""
+    deletecolmetadata!(x, [col], [key::AbstractString])
+
+Delete metadata for table `x` for column `col` for key `key` and return `x`
+(if metadata for `key` is not present do not perform any action).
+If `key` is not passed delete all metadata for table `x` for column `col`.
+If only `x` is passed delete all column level metadata for table `x`.
+If `x` does not support metadata deletion for column `col` throw `ArgumentError`.
+"""
+deletecolmetadata!(::T, :Symbol, ::AbstractString) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
+deletecolmetadata!(::T, :Int, ::AbstractString) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
+deletecolmetadata!(::T, :Symbol) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
+deletecolmetadata!(::T, :Int) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
+deletecolmetadata!(::T) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -419,11 +419,11 @@ If `x` does not support metadata deletion for column `col` throw `ArgumentError`
 """
 deletecolmetadata!(::T, ::Symbol, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletecolmetadata!(::T, :Int, ::AbstractString) where {T} =
+deletecolmetadata!(::T, ::Int, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 deletecolmetadata!(::T, ::Symbol) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletecolmetadata!(::T, :Int) where {T} =
+deletecolmetadata!(::T, ::Int) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 deletecolmetadata!(::T) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -391,8 +391,8 @@ preferred behavior if it is possible) or returns `()` (this duality is allowed
 as some Tables.jl tables do not have a schema).
 
 If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
-pairs for all columns that have metadata. If `x` does not support column
-metadata return `()`.
+pairs for all columns that have metadata, where `col` are `Symbol`.
+If `x` does not support column metadata return `()`.
 """
 colmetadatakeys(::Any, ::Int) = ()
 colmetadatakeys(::Any, ::Symbol) = ()
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -332,7 +332,7 @@ If `x` does not support setting metadata throw `ArgumentError`.
 
 $STYLE_INFO
 """
-metadata!(::T, ::Any; style) where {T} =
+metadata!(::T, ::Any, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
 
 """
@@ -380,7 +380,7 @@ If `x` does not support setting metadata for column `col` throw `ArgumentError`.
 
 $STYLE_INFO
 """
-colmetadata!(::T, ::Any, ::Any; style) where {T} =
+colmetadata!(::T, ::Any, ::Any, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -367,7 +367,7 @@ col, key)` returns a metadata value. If `x` does not support metadata for column
 
 `col` must have a type that is supported by table `x` for column indexing.
 Following the Tables.jl contract `Symbol` and `Int` are always allowed. Passing
-`col` that is not a column of `x` can either throws an error (this is a
+`col` that is not a column of `x` either throws an error (this is a
 preferred behavior if it is possible) or returns `()` (this duality is allowed
 as some Tables.jl tables do not have a schema).
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -287,4 +287,11 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+"""
+    getmetadata(x, ...)
+
+Return metadata associated with object `x` or `nothing` if it doest not support metadata.
+"""
+getmetadata(::Any) = nothing
+
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -288,34 +288,67 @@ using a `sink` function to materialize the table.
 function allcombinations end
 
 """
-    metadata(x, [key])
+    metadata(x)
 
 Return metadata associated with object `x` as an `AbstractDict{String}` object
 (or an object implementing the same interface).
-
-If the optional `key` argument is passed, return metadata
-associated with the indicated key (this feature can be used, for example,
-to return metadata attached to a given column in a Tables.jl table).
 
 Note that some systems, like Arrow.jl, might assume that metadata values are
 also `String`.
 """
 metadata(::T) where {T} =
     throw(ArgumentError("Metadata is currently not supported for values of type $T"))
-metadata(::T, ::Any) where {T} =
+
+"""
+    colmetadata(table, column)
+
+Return metadata associated with column `column` of Tables.jl table `table` as an
+`AbstractDict{String}` object (or an object implementing the same interface).
+
+Note that some systems, like Arrow.jl, might assume that metadata values are
+also `String`.
+"""
+colmetadata(::T, ::Any) where {T} =
     throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
 
 """
-    hasmetadata(x, [key])
+    colmetadata(table)
+
+Return metadata associated with columns of Tables.jl table `table` as an
+`AbstractDict{<:Any, <:AbstractString{String}}` object (or an object implementing
+the same interface).
+
+The returned dictionary must contain columns `column` for which
+`hascolmetadata(table, column)` returns `true` and a value associated to such
+key must be `colmetadata(table, column)`.
+"""
+colmetadata(::T, ::Any) where {T} =
+    throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
+
+"""
+    hasmetadata(x)
 
 Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
 Return `nothing` if `x` does not support metadata.
-
-If the optional `key` argument is passed, return information if metadata
-is associated with `key` (this feature can be used, for example,
-to query about metadata attached to a given column in a Tables.jl table).
 """
 hasmetadata(::Any) = nothing
-hasmetadata(::Any, ::Any) = nothing
+
+"""
+    hascolmetadata(table, column)
+
+Return `true` if column `column` of Tables.jl table `table` has non-empty metadata,
+and return `false` if metadata is empty.
+Return `nothing` if metadata is not supported.
+"""
+hascolmetadata(::Any, ::Any) = nothing
+
+"""
+    hascolmetadata(table)
+
+Return `true` if at least one column of of Tables.jl table `table` has non-empty
+metadata, and return `false` if metadata is empty for all columns.
+Return `nothing` if metadata is not supported.
+"""
+hascolmetadata(::Any) = nothing
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -311,8 +311,8 @@ metadata(::T, ::Any) where {T} =
 Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
 Return `nothing` if `x` does not support metadata.
 
-If the optional `key` argument is passed, return information about whether
-metadata is associated with the indicated key (this feature can be used, for example,
+If the optional `key` argument is passed, return information if metadata
+is associated with `key` (this feature can be used, for example,
 to query about metadata attached to a given column in a Tables.jl table).
 """
 hasmetadata(::Any) = nothing
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -288,13 +288,33 @@ using a `sink` function to materialize the table.
 function allcombinations end
 
 """
-    metadata(x)
+    metadata(x, [column])
 
 Return metadata associated with object `x` as an `AbstractDict{String}` object
-(or an object implementing the same interface), or `nothing` if `x` does not support metadata.
+(or an object implementing the same interface).
+
+Optionally for Tables.jl tables a `columns` argument can be passed, in which case
+return metadata associated with the indicated column.
 
 Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
 """
-metadata(::Any) = nothing
+metadata(::T) where {T} = throw(ArgumentError("Metadata is currently not " *
+                                              "supported for values of type $T"))
+metadata(::T, ::Any) where {T} = throw(ArgumentError("Column metadata is currently not " *
+                                                     "supported for values of type $T"))
+
+"""
+    hasmetadata(x, [column])
+
+Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
+Return `nothing` if `x` does not support metadata.
+
+Optionally for Tables.jl tables a `columns` argument can be passed, in which case
+return the same information about metadata associated with the indicated column.
+
+Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
+"""
+hasmetadata(::Any) = nothing
+hasmetadata(::Any, ::Any) = nothing
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -379,7 +379,12 @@ If `col` is passed return an iterator of metadata keys for which
 `metadata(x, col, key)` returns a metadata value.
 If `x` does not support metadata for column `col` return `()`.
 
-$COL_INFO
+`col` must be a type that is supported by table `x`. Following the Tables.jl
+contract `Symbol` and `Int` must be supported. However, tables that allow other
+column indexing (e.g., using strings or integers other than `Int`) are allowed
+to accept such types of `col` argument. Passing `col` that is not a column
+of `x` can either throw an error or return `()` (this duality is allowed as
+some Tables.jl do not have schema).
 
 If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
 pairs for all columns that have metadata.
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -288,29 +288,32 @@ using a `sink` function to materialize the table.
 function allcombinations end
 
 """
-    metadata(x, [column])
+    metadata(x, [key])
 
 Return metadata associated with object `x` as an `AbstractDict{String}` object
 (or an object implementing the same interface).
 
-Optionally for Tables.jl tables a `columns` argument can be passed, in which case
-return metadata associated with the indicated column.
+Optionally `key` argument can be passed, in which case return metadata
+associated with the indicated key (this feature can be used, for example,
+by Tables.jl table to return metadata attached to columns of a table).
 
-Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
+Note that some systems, like Arrow.jl, might assume that metadata values are
+also `String`.
 """
 metadata(::T) where {T} = throw(ArgumentError("Metadata is currently not " *
                                               "supported for values of type $T"))
-metadata(::T, ::Any) where {T} = throw(ArgumentError("Column metadata is currently not " *
+metadata(::T, ::Any) where {T} = throw(ArgumentError("Key metadata is currently not " *
                                                      "supported for values of type $T"))
 
 """
-    hasmetadata(x, [column])
+    hasmetadata(x, [key])
 
 Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
 Return `nothing` if `x` does not support metadata.
 
-Optionally for Tables.jl tables a `columns` argument can be passed, in which case
-return the same information about metadata associated with the indicated column.
+Optionally `key` argument can be passed, in which case return information if
+metadata is associated with the indicated key (this feature can be used, for example,
+by Tables.jl table to query about metadata attached to columns of a table).
 
 Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
 """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -297,7 +297,7 @@ Note that some systems, like Arrow.jl, might assume that metadata values are
 also `String`.
 """
 metadata(::T) where {T} =
-    throw(ArgumentError("Metadata is currently not supported for values of type $T"))
+    throw(ArgumentError("Objects of type $T do not support metadata"))
 
 """
     colmetadata(table, column)
@@ -309,7 +309,7 @@ Note that some systems, like Arrow.jl, might assume that metadata values are
 also `String`.
 """
 colmetadata(::T, ::Any) where {T} =
-    throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
+    throw(ArgumentError("Objects of type $T do not support column metadata"))
 
 """
     colmetadata(table)
@@ -323,7 +323,7 @@ The returned dictionary must contain columns `column` for which
 key must be `colmetadata(table, column)`.
 """
 colmetadata(::T) where {T} =
-    throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
+    throw(ArgumentError("Objects of type $T do not support column metadata"))
 
 """
     hasmetadata(x)
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -305,6 +305,8 @@ metadata(::T) where {T} =
 Return metadata associated with column `column` of Tables.jl table `table` as an
 `AbstractDict{String}` object (or an object implementing the same interface).
 
+`column` should be any accepted column identifier by `table`, but at least `Symbol` must be supported.
+
 Note that some systems, like Arrow.jl, might assume that metadata values are
 also `String`.
 """
@@ -328,6 +330,9 @@ hasmetadata(::Any) = nothing
 Return `true` if column `column` of Tables.jl table `table` has non-empty metadata,
 and return `false` if metadata is empty.
 Return `nothing` if metadata is not supported for `column`.
+
+
+`column` should be any accepted column identifier by `table`, but at least `Symbol` must be supported.
 
 If `hascolmetadata` returns `Bool` then it is guaranteed that call to
 `colmetatada(table, columns)` will not throw an error.
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -291,7 +291,8 @@ const STYLE_INFO = """
 One of the uses of the metadata `style` is decision
 how the metadata should be propagated when `x` is transformed. This interface
 defines the `:none` style that indicates that metadata should not be propagated
-under transformations. All types supporting metadata allow at least this style.
+under any operations (it is only preserved when a copy of the source table is
+performed). All types supporting metadata allow at least this style.
 """
 
 const COL_INFO = """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -338,7 +338,7 @@ metadata!(::T, ::AbstractString, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
 
 """
-    deletemetadata!(x, [key::AbstractString])
+    deletemetadata!(x, key::AbstractString)
 
 Delete metadata for object `x` for key `key` and return `x`
 (if metadata for `key` is not present do not perform any action).
@@ -347,7 +347,14 @@ If `x` does not support metadata deletion throw `ArgumentError`.
 """
 deletemetadata!(::T, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletemetadata!(::T) where {T} =
+
+"""
+    emptymetadata!(x)
+
+Delete all metadata for object `x`.
+If `x` does not support metadata deletion throw `ArgumentError`.
+"""
+emptymetadata!(::T) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 
 """
@@ -409,23 +416,29 @@ colmetadata!(::T, ::Symbol, ::AbstractString, ::Any; style) where {T} =
     throw(ArgumentError("Objects of type $T do not support setting metadata"))
 
 """
-    deletecolmetadata!(x, [col], [key::AbstractString])
+    deletecolmetadata!(x, col, key::AbstractString)
 
 Delete metadata for table `x` for column `col` for key `key` and return `x`
 (if metadata for `key` is not present do not perform any action).
-If `key` is not passed delete all metadata for table `x` for column `col`.
-If only `x` is passed delete all column level metadata for table `x`.
 If `x` does not support metadata deletion for column `col` throw `ArgumentError`.
 """
 deletecolmetadata!(::T, ::Symbol, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 deletecolmetadata!(::T, ::Int, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletecolmetadata!(::T, ::Symbol) where {T} =
+
+"""
+    emptycolmetadata!(x, [col])
+
+Delete all metadata for table `x` for column `col`.
+If `col` is not passed delete all column level metadata for table `x`.
+If `x` does not support metadata deletion for column `col` throw `ArgumentError`.
+"""
+emptycolmetadata!(::T, ::Symbol) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletecolmetadata!(::T, ::Int) where {T} =
+emptycolmetadata!(::T, ::Int) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletecolmetadata!(::T) where {T} =
+emptycolmetadata!(::T) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -312,20 +312,6 @@ colmetadata(::T, ::Any) where {T} =
     throw(ArgumentError("Objects of type $T do not support column metadata"))
 
 """
-    colmetadata(table)
-
-Return metadata associated with columns of Tables.jl table `table` as an
-`AbstractDict{<:Any, <:AbstractString{String}}` object (or an object implementing
-the same interface).
-
-The returned dictionary must contain columns `column` for which
-`hascolmetadata(table, column)` returns `true` and a value associated to such
-key must be `colmetadata(table, column)`.
-"""
-colmetadata(::T) where {T} =
-    throw(ArgumentError("Objects of type $T do not support column metadata"))
-
-"""
     hasmetadata(x)
 
 Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
@@ -347,17 +333,5 @@ If `hascolmetadata` returns `Bool` then it is guaranteed that call to
 `colmetatada(table, columns)` will not throw an error.
 """
 hascolmetadata(::Any, ::Any) = nothing
-
-"""
-    hascolmetadata(table)
-
-Return `true` if at least one column of of Tables.jl table `table` has non-empty
-metadata, and return `false` if metadata is empty for all columns.
-Return `nothing` if metadata is not supported.
-
-If `hascolmetadata` returns `Bool` then it is guaranteed that call to
-`colmetatada(table)` will not throw an error.
-"""
-hascolmetadata(::Any) = nothing
 
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -417,11 +417,11 @@ If `key` is not passed delete all metadata for table `x` for column `col`.
 If only `x` is passed delete all column level metadata for table `x`.
 If `x` does not support metadata deletion for column `col` throw `ArgumentError`.
 """
-deletecolmetadata!(::T, :Symbol, ::AbstractString) where {T} =
+deletecolmetadata!(::T, ::Symbol, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 deletecolmetadata!(::T, :Int, ::AbstractString) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
-deletecolmetadata!(::T, :Symbol) where {T} =
+deletecolmetadata!(::T, ::Symbol) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
 deletecolmetadata!(::T, :Int) where {T} =
     throw(ArgumentError("Objects of type $T do not support metadata deletion"))
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -322,7 +322,7 @@ The returned dictionary must contain columns `column` for which
 `hascolmetadata(table, column)` returns `true` and a value associated to such
 key must be `colmetadata(table, column)`.
 """
-colmetadata(::T, ::Any) where {T} =
+colmetadata(::T) where {T} =
     throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
 
 """
@@ -330,6 +330,9 @@ colmetadata(::T, ::Any) where {T} =
 
 Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
 Return `nothing` if `x` does not support metadata.
+
+If `hasmetadata` returns `Bool` then it is guaranteed that call to `metatada(x)`
+will not throw an error.
 """
 hasmetadata(::Any) = nothing
 
@@ -338,7 +341,10 @@ hasmetadata(::Any) = nothing
 
 Return `true` if column `column` of Tables.jl table `table` has non-empty metadata,
 and return `false` if metadata is empty.
-Return `nothing` if metadata is not supported.
+Return `nothing` if metadata is not supported for `column`.
+
+If `hascolmetadata` returns `Bool` then it is guaranteed that call to
+`colmetatada(table, columns)` will not throw an error.
 """
 hascolmetadata(::Any, ::Any) = nothing
 
@@ -348,6 +354,9 @@ hascolmetadata(::Any, ::Any) = nothing
 Return `true` if at least one column of of Tables.jl table `table` has non-empty
 metadata, and return `false` if metadata is empty for all columns.
 Return `nothing` if metadata is not supported.
+
+If `hascolmetadata` returns `Bool` then it is guaranteed that call to
+`colmetatada(table)` will not throw an error.
 """
 hascolmetadata(::Any) = nothing
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -288,19 +288,17 @@ using a `sink` function to materialize the table.
 function allcombinations end
 
 const STYLE_INFO = """
-One of the uses of the metadata style information is decision
+One of the uses of the metadata `style` is decision
 how the metadata should be propagated when `x` is transformed. This interface
 defines the `:none` style that indicates that metadata should not be propagated
-under transformations. At least this style must be supported by any type
+under transformations. At least this style is allowed by any type
 defining support for metadata.
 """
 
 const COL_INFO = """
-`col` must be a type that is supported by table `x`. Following the Tables.jl
-contract `Symbol` and `Int` must be supported. However, tables that allow other
-column indexing (e.g., using strings or integers other than `Int`) are allowed
-to accept such types of `col` argument. Passing `col` that is not a column
-of `x` must throw an error.
+`col` must have a type that is supported by table `x` for column indexing.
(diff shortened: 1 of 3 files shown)


src/DataAPI.jl
```diff
@@ -287,4 +287,11 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+"""
+    getmetadata(x, ...)
```
nalimilan, line 291: How about calling this just `metadata`? The only `get*` methods that exist in Base are those to retrieve a particular element (`getindex`, `getkey`...).

Also, why not allow only a single argument for now? If we want, we can allow passing a second argument giving the name of the metadata to retrieve, but in that case better specify that it should be a string (or a symbol?.)

src/DataAPI.jl
```diff
@@ -287,4 +287,11 @@ using a `sink` function to materialize the table.
...
+
+Return metadata associated with object `x` or `nothing` if it doest not support metadata.
```
nalimilan, line 293: Let's give more guarantees?
```suggestion
Return metadata associated with object `x` as an `AbstractDict{String,Any}` object
(or an object implementing the same interface), or `nothing` if `x` does not support metadata.
```

I think requiring `AbstractDict{String,String}` would be too strict as it can be useful to store e.g. the observation date, the duration of observation, a numeric index, the parent population size...

src/DataAPI.jl
bkamins: We could call it `metadata`, but the question is if @quinnj would then agree to rename the respective function in Arrow.jl for consistency.

bkamins: > Also, why not allow only a single argument for now?

We could, but I just wanted to signal that `metadata` would dispatch on a first positional argument and that specific types are allowed to opt-in with passing more than one argument if needed. This flexibility is not needed in DataFrames.jl now, but might be needed in the future when we would pass column name or column number as a second argument.

src/DataAPI.jl
bkamins: Yes - it would be useful. We can make this more flexible. The only thing is that Arrow.jl now only accepts strings, so the question is if this can be lifted there? The point is that I wanted, in particular, to ensure that if in DataFrames.jl we set some metadata it can be stored and retrieved by Arrow.jl.

src/DataAPI.jl
nalimilan: > We could call it metadata, but the question is if @quinnj would then agree to rename the respective function in Arrow.jl for consistency.

OK, I hadn't realized that this was the Arrow.jl name. Let's hear what @quinnj thinks.

> We could, but I just wanted to signal that metadata would dispatch on a first positional argument and that specific types are allowed to opt-in with passing more than one argument if needed. This flexibility is not needed in DataFrames.jl now, but might be needed in the future when we would pass column name or column number as a second argument.

Then I would only document the method we expect packages to implement for now, and add more methods later if needed. Having `...` in signatures isn't super useful as it doesn't allow consumers to assume anything about the API.

src/DataAPI.jl
nalimilan: There's a tension here. I'd favor allowing any element type as it would be too bad to limit what we can do in Julia just because Arrow doesn't support it. This would be similar to allowing only element types supported by Arrow in data frame columns. :-) Maybe Arrow.jl can call `string` on each element before saving it?

src/DataAPI.jl
quinnj: I like just `metadata`

src/DataAPI.jl
quinnj: The Arrow.jl `{String, String}` restriction is strict, but I agree that we could just call `string` on the values to convert (I think we already do this in fact). I think we could maybe mention that some systems may assume `Dict{String, String}`, and so it's suggested but not a hard requirement? Something like that?

quinnj (approved):
Sorry for the slow review here; (getting ramped up in the new job and trying to get caught up a on a lot of Julia stuff + prioritize for the future). I like the name change to just `metadata` and getting this merged. I can make the required changes in Arrow.jl.

src/DataAPI.jl
bkamins, line 291:
```suggestion
    metadata(x)
```

src/DataAPI.jl
```diff
@@ -287,4 +287,11 @@ using a `sink` function to materialize the table.
...
+"""
+getmetadata(::Any) = nothing
```
bkamins, line 295:
```suggestion
metadata(::Any) = nothing
```

src/DataAPI.jl
```diff
@@ -287,4 +287,12 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+"""
+    metadata(x)
+
+Return metadata associated with object `x` as an `AbstractDict{String}` object
+(or an object implementing the same interface), or `nothing` if `x` does not support metadata.
```
bkamins, line 294:
```suggestion
(or an object implementing the same interface), or `nothing` if `x` does not support metadata.

Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
```

src/DataAPI.jl
bkamins: OK - I have updated the PR following the comments.
@nalimilan - could you please, in particular, check wording of the docstring. Thank you!

test/runtests.jl
```diff
@@ -173,4 +174,9 @@ end
     @test DataAPI.unwrap(missing) === missing
 end
 
+@testset "metadata" begin
+    @test isnothing(DataAPI.metadata(1))
+    @test DataAPI.metadata(TestArray([1, 2])) === Dict("length => 2)
```
bkamins, line 179:
```suggestion
    @test DataAPI.metadata(TestArray([1, 2])) === Dict("length" => 2)
```

test/runtests.jl
```diff
@@ -173,4 +174,9 @@ end
     @test DataAPI.unwrap(missing) === missing
 end
 
+@testset "metadata" begin
+    @test isnothing(DataAPI.metadata(1))
+    @test DataAPI.metadata(TestArray([1, 2])) === Dict("length" => 2)
```
bkamins, line 179:
```suggestion
    @test DataAPI.metadata(TestArray([1, 2])) == Dict("length" => 2)
```

test/runtests.jl
```diff
@@ -173,4 +174,9 @@ end
     @test DataAPI.unwrap(missing) === missing
 end
 
+@testset "metadata" begin
+    @test isnothing(DataAPI.metadata(1))
```
bkamins, line 178:
```suggestion
    @test DataAPI.metadata(1) === nothing
```

nalimilan (approved):
Looks good! Though we may want to hold this until we have an implementation ready in DataFrames.jl and/or in Tables.jl just in case we discover unanticipated needs.

bkamins:
I will do implementation in DataFrames.jl then and I understand that @quinnj will do implementation in Arrow.jl - right?
I guess these would be "pending PRs" waiting for this to be merged and released, but at least we would see if there were no implementation issues.

@nalimilan - What would you want implemented in Tables.jl?

bkamins:
I have started the implementation and see one problem. I propose to discuss it here (although it is DataFrames.jl specific).

Assume that I want to add metadata to some data frame that does not have metadata yet. I run `medatada(df)` and I get `nothing`. Now there is no way to add metadata to `nothing`. So, while I am OK to return `nothing` by default, I think that in DataFrames.jl we should always return `Dict{String}`. The only thing I would do is that creation of this dictionary would be lazy - i.e. by default it would be `nothing`, but if user called `metadata` then the dictionary would be created on a first call. Technically then this should be `metadata!` as it would mutate the source data frame, but I think it is OK, to call this function `metadata` and hide the fact that the creation of the dictionary is delayed.

@nalimilan - is this OK for you?

nalimilan:
Yeah it sounds fine to create the dict on the first call to avoid returning `nothing`.

BTW, instead of recommending users mutate the `Dict` directly we could have a `metadata!` function that could give a somewhat simpler API for basic operations, with e.g. `metadata!(df, "month" => "January")` instead of `metadata(df)["month"] = "January"` (which looks awfully like setting dimension names in R). But probably better start with a minimal API for now.

bkamins:
I started implementing it and it seems that we need `metadata!` the way you proposed it. The reason is that we need to return `nothing` if there is no metadata in a data frame. Otherwise there would be no way to check that some object does not have metadata (i.e. checking for metadata would always create an empty dictionary which could be unwanted).

The `metadata!` signature should be:
```
metadata!(table, pair; mode::Symbol=:overwrite)
```
where `mode` would be:
* `:overwrite` -> new metadata overwrites old one if it exists
* `:leave` -> if metadata existed it is unchanged (i.e. new value is ignored)
* `:error` -> throw an error if existing key is tried to be assigned

Side note: to be able to add metadata to `DataFrame` we have to make it `mutable struct` (now it is `struct` so we could not change the value from `nothing` to dictionary).

@nalimilan - do you have any additional thoughts on these two points?

bkamins:
In https://github.com/JuliaData/DataFrames.jl/pull/3055 I have drafted the implementation so you can see the important aspects of the design.
(I have made the draft minimal - including only essential parts so that it is easy to review at this stage)

nalimilan:
Indeed I also realized that. :-/

Now that you list the requirements, `metadata!` isn't super appealing as a general solution. It's nice for simple things like `metadata!(df, "month" => "January")`, but having to add a `mode` argument amounts to reinventing the `Dict` API, forcing users to learn a new way to what is basically modifying a `Dict`. And that doesn't even include a syntax to remove a key or remove all metadata...

One alternative would be to have an `hasmetadata` function [like in Metadata.jl](https://tokazama.github.io/Metadata.jl/stable/#Metadata.has_metadata) that could be used to check whether there's any metadata instead of calling `metadata`.

Cc: @Tokazama

src/DataAPI.jl
```diff
@@ -287,4 +287,14 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+"""
+    metadata(x)
```
bkamins, line 291:
```suggestion
    metadata(x, [column])
```

src/DataAPI.jl
```diff
@@ -287,4 +287,14 @@ using a `sink` function to materialize the table.
...
+
+Return metadata associated with object `x` as an `AbstractDict{String}` object
+(or an object implementing the same interface), or `nothing` if `x` does not support metadata.
```
bkamins, line 294:
```suggestion
(or an object implementing the same interface), or `nothing` if `x` does not support metadata.

If `x` is a Tables.jl table then optionally a second argument `column` can be passed,
in which case the dictionary with metadata attached to this column is returned
or `nothing` if column has no attached metadata.
```

src/DataAPI.jl
```diff
@@ -287,4 +287,14 @@ using a `sink` function to materialize the table.
...
+
+Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
+"""
+metadata(::Any) = nothing
```
bkamins, line 298:
```suggestion
metadata(::Any) = nothing
metadata(::Any, ::Any) = nothing
```

nalimilan:
> Are we assuming that metadata always refers to a mutable field? With arrays we typically want to wrap an array and metadata in a new type.

@Tokazama Can you develop what you mean by "refers to a mutable field"? The idea of this PR is that `metadata(obj)` would return an `AbstractDict` object which could be mutable if the object type supports it (e.g. for `DataFrame`) but not necessarily. This works with array wrappers AFAICT.

Tokazama:
It appears that the assumption is that the object

1. Has a `metadata` field that has metadata stored
2. Has a `metadata` field with no metadata stored
3. Doesn't support metadata storage

This assumes we can only attach metadata to something that has an existing field dedicated to metadata that is mutable.

nalimilan:
Why do you mention a "field"? Types can implement the method however they want, they don't need to have a dedicated field. Actually, my thinking was that a package like Metadata.jl could define a fallback method `metadata(::Any)` once it includes a general mechanism to store metadata in a global dict. (This would require defining in DataAPI who owns this generic definition.)

bkamins:
@Tokazama - or to state what @nalimilan has written more explicitly.

In DataAPI.jl we want to define API that does not assume anything. It should be only an interface that is maximally generic with the only comment that sicne it is "Data" API we want it to be aware of Tables.jl tables (for convenience, see bullet 5 below).

The logic is the following:
1. To check if some object supports metadata call `hasmetadata`, the reason why we provide 3 options of return value is that we need to distinguish three cases (note that we do not state how this metadata should be stored - we only ask if object has metadata):
    * `true` - object supports metadata and already has some metadata stored;
    * `false` - object supports metadata but no metadata is stored yet
    * `nothin` - object does not support metadata
3. If you know that `hasmetadata` returns `Bool` then you can ask the object to return its dictionary-like metadata with keys being `String`. We do not specify how this data should be stored in the object (or if it should be stored at all - the returned dictionary-like object can be generated lazily when the call is made)
4. We do not say if the returned dict-like object should be mutable or not. E.g. for Arrow.jl it is immutable, for DataFrames.jl it will be mutable. The difference is that if a mutable object is returned you can use it to add/remove metadata from an object by mutating it in-place. If immutable object is returned this means that object has metadata but this metadata cannot be changed by the user.
5. One case that we need to handle is that users might want to add metadata to objects that did not think of providing metadata on their own. For example a static array or a range can be a column of a data frame. Since static array or range does not provide a way to store metadata in it, and also it is immutable, which means that other ways to attach metadata to such objects are not possible, we provide a workaround approach, in which we say that "wrapper table object" can store metadata for such a column, and you can query this wrapper for metadata of any column (including a column that technically does not support metadata).
6. Indeed there are some column types that support metadata. For example this is the case in Arrow.jl. Therefore I assume that in Arrow.jl @quinnj will implement something like (this is pseudocode) `metadata(arrowtable, arrowcolumn) = metadata(getcolumn(arrowtable, arrowcolumn))` (the same for `hasmetadata`) and all works as expected.

@Tokazama + @quinnj What do you think about it?

Tokazama:
>  in this way if the user does not need metadata we have a better performance; or e.g. object might register itself for metadata in Metadata.jl only if user asks for it, but if user does not ask for it then just false is returned meaning that object "can" register to Metadata.jl metadata but has not done it yet)

So something like `hasmetadata(::AbstractArray) = false` could still be performed? That's fine with me.

bkamins:
> So something like `hasmetadata(::AbstractArray) = false` could still be performed?

Yes, but not `AbstractArray` but:
```
hasmetadata(::SomeSpecificMutableAbstractArray) = false
```
The reason is that in general there is no way to attach metadata to e.g. `1:3` range since it is immutable (but you can attach metadata to it if it were a column in a table if table in which it were stored supports metadata).

bkamins:
> `metadata(::Any)` shouldn't search for globally stored metadata by default

It does not, by default it is an error (see the default implementation in this PR). It will only search for metadata if `metadata(::SomeSpecificType)` method is defined by the user/some package.

Tokazama:
> > `metadata(::Any)` shouldn't search for globally stored metadata by default
> 
> It does not, by default it is an error (see the default implementation in this PR). It will only search for metadata if `metadata(::SomeSpecificType)` method is defined by the user/some package.

I thought the implication was that literally it be defined any `Any` in Metadata.jl

bkamins:
> I thought the implication was that literally it be defined any `Any` in Metadata.jl

It cannot be, as it would be type piracy. Default implementation for `Any` is to throw an error (metadata is not supported). Then specific types can opt-in to support metadata. In particular such types can delegate handling of their metadata to Metadata.jl. However, we cannot require Metdata.jl as dependency of DataAPI.jl as there might be cases that users want to depend on DataAPI.jl but do not want Metadata.jl as dependency (e.g. they know they do not want metadata and want to minimize the list of packages they depend on).

Tokazama:
Sorry for the confusion. The only other thing I might suggest is that the language be changed from being specific to columns. It's not a big deal, but I assume `metadata(x, key)` is would be broadly useful and that only `DataFrame` would specifically be referring to columns.

bkamins:
sure `key` makes sense. I am changing it now.

(for a reference - there are multiple Tables.jl tables implementation for which `key` would be column, but of course it is not a problem)

Tokazama:
Thanks! Once this goes through and the shake ups with `ArrayInterface` are through I'll update Metadata.jl to use this.

bkamins:
@pdeffebach - can you please review. Thank you!

nalimilan (commented):
Great. Just a few wording suggestions then.

quinnj:
Yes, I think this sounds like a good plan. My biggest question that I think I just need to solve on the Arrow.jl side is how to properly handle the `copy(::Arrow.Table)` operation; the functionality is to "materialize" a mutable copy of the immutable arrow data. Until now, we've just ignored metadata, but it would be nice to include any metadata, either table or column, in the materialized table.

Tokazama:
> Yes, I think this sounds like a good plan. My biggest question that I think I just need to solve on the Arrow.jl side is how to properly handle the `copy(::Arrow.Table)` operation; the functionality is to "materialize" a mutable copy of the immutable arrow data. Until now, we've just ignored metadata, but it would be nice to include any metadata, either table or column, in the materialized table.

I've found it really difficult to know what should be propagated with `copy` and `similar`. I'm not sure there's a right answer to that one so it might have to be a case by case basis for the data type outside of a standard API.

bkamins:
> but it would be nice to include any metadata, either table or column, in the materialized table.

In DataFrames.jl I plan that `copy` will retain both table and column metadata. However, I think, as @Tokazama, that it will have to be on a case-by-case basis. DataAPI.jl API does not introduce any requirements on "metadata propagation", "metadata updating" or "metadata storage" it is only meant to standardize "metadata format and discovery" across the ecosystem.

What I want is that if e.g. you take `DataFrame` and save it as Arrow file then table and column metadata should be retained (subject to conversion of values to strings). Similarly if you convert Arrow table to `DataFrame` then table and column metadata should be retained.

src/DataAPI.jl
```diff
@@ -287,4 +287,37 @@ using a `sink` function to materialize the table.
...
+    hasmetadata(x, [key])
+
+Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
+Return `nothing` if `x` does not support metadata.
+
+If the optional `key` argument is passed, return information about whether
+metadata is associated with the indicated key (this feature can be used, for example,
+to query about metadata attached to a given column in a Tables.jl table).
+
+Note that some systems, like Arrow.jl, might assume that metadata values are also `String`.
```
bkamins, line 318:
```suggestion
```

src/DataAPI.jl
```diff
@@ -287,4 +287,35 @@ using a `sink` function to materialize the table.
...
+    throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
+
+"""
+    hasmetadata(x, [key])
+
+Return `true` if `x` has non-empty metadata, and return `false` if metadata is empty.
+Return `nothing` if `x` does not support metadata.
+
+If the optional `key` argument is passed, return information about whether
+metadata is associated with the indicated key (this feature can be used, for example,
```
bkamins, line 315:
```suggestion
If the optional `key` argument is passed, return information if metadata
is associated with `key` (this feature can be used, for example,
```

Tokazama:
I have some ideas of how to manage propagation of metadata in a more structured way but it would probably be better for me to bring those up in a subsequent PR rather than bog this down with potentially unhelpful suggestions.

bkamins:
> I have some ideas of how to manage propagation of metadata

This would be great to discuss as this is the biggest issue in DataFrames.jl implementation of metadata. However, as you commented, it is probably better to open another thread for this. Thank you!

nalimilan:
Maybe continue the discussion at https://github.com/JuliaData/DataAPI.jl/issues/22?

nalimilan:
I've just noticed that FileIO exports a `metadata` function that Images.jl and ImageMagick.jl use to allow accessing some information without loading a whole file. See https://github.com/JuliaIO/FileIO.jl/pull/178 and linked PRs. Unfortunately, that function is undocumented (it's not listed in the docs at all), so implementations are allowed to return anything, and the current implementations appear to return a tuple.

We should probably ask them whether they would be OK to import `DataAPI.metadata` to avoid name collisions (which already happen e.g. with RData). But I'm not sure how to handle the API conflict. Maybe Images.jl and ImageMagick.jl could move to returning a `NamedTuples`, which is backward-compatible with and more informative than a tuple, and a bit more similar to `AbstractDict{String}`? Though they would still not really implement the DataAPI definition as this PR documents it.

Tokazama:
I don't think they reexport everything from FileIO https://github.com/JuliaImages/Images.jl/blob/401d1465196bec298d2035a6c625af78244f04e7/src/Images.jl#L13.

nalimilan:
Ah, indeed. Though any package reexporting FileIO also exports `metadata` even if it doesn't use it (this is the case with RData.jl).

src/DataAPI.jl
```diff
@@ -287,4 +287,35 @@ using a `sink` function to materialize the table.
...
+associated with the indicated key (this feature can be used, for example,
+to return metadata attached to a given column in a Tables.jl table).
+
+Note that some systems, like Arrow.jl, might assume that metadata values are
+also `String`.
+"""
+metadata(::T) where {T} =
+    throw(ArgumentError("Metadata is currently not supported for values of type $T"))
+metadata(::T, ::Any) where {T} =
+    throw(ArgumentError("Key metadata is currently not supported for values of type $T"))
```
nalimilan, line 306: This fallback can give misleading errors if you pass the wrong argument. For example:
```julia
julia> metadata(DataFrame(x=1), 1.0)
ERROR: ArgumentError: Key metadata is currently not supported for values of type DataFrame
```

Maybe something like this would be clearer?
```suggestion
metadata(::T) where {T} =
    throw(ArgumentError("Objects of type $T do not support metadata"))
metadata(::T, ::U) where {T, U} =
    throw(ArgumentError("Objects of type $T do not support metadata for keys of type $U"))
```

Tokazama: Would we be assuming too much if we just did `metadata(x, key) = metadata(x)[key]`? Then we'd fall back to the appropriate error if there's no metadata and get the appropriate error of indexing by key isn't supported.

bkamins: `metadata(x, key)` is a metadata dictionary associated with `key`,
while
`meatadata(x)[key]` is a value associated with `key` key for metadata dictionary associated with `x`.

In short `metadata(x, key)` and `metadata(x)` are separate metadata dictionaries (here `key` serves as a kind-of additional metadata scope specifier).

Tokazama: Oh, so you would use this to do something like `metadata(x, :column)`?

bkamins: Yes - in Tables.jl table context. However, I assume that in different contexts `key` can mean something different.

alyst: I think `metadata(frame, column)` API would be a constant source of confusion with people assuming it is `metadata(frame, key)`.
What about `colmetadata(frame, column)` or allowing `metadata(frame.column)`?
The latter, of course, would require that `frame.column` returns some thin wrapper around column data, which is quite a change.

bkamins: > `frame.column` returns some thin wrapper around column data

The whole point of this API extension is that this thin wrapper is very problematic and would confuse users in other contexts.

`colmetadata` is OK I think, then also `colmetadata(frame)` could return a collection of all column metadata defined for the whole table. @nalimilan - what do you think?

nalimilan: Good point. I was also concerned that `metadata(df, col)` isn't consistent with `getindex(df, :, col)`. Using `colmetadata` is more explicit and it avoids the possibility of conflicts between different interfaces.

Tokazama: Since it came up, my recent comment [here](https://github.com/JuliaData/DataAPI.jl/issues/22#issuecomment-1137755587) may be of interest. When the context we assign to metadata (what it is attached to) is an in memory object there's a reasonably small grouping we can create for referring to metadata and its relationship to data. I'm not sure how much more you care to add to this PR or to subsequent PRs.

src/DataAPI.jl
```diff
@@ -288,34 +288,76 @@ using a `sink` function to materialize the table.
...
+`AbstractDict{<:Any, <:AbstractString{String}}` object (or an object implementing
+the same interface).
 
+The returned dictionary must contain columns `column` for which
+`hascolmetadata(table, column)` returns `true` and a value associated to such
+key must be `colmetadata(table, column)`.
 """
-    hasmetadata(x, [key])
+colmetadata(::T) where {T} =
+    throw(ArgumentError("Objects of type $T do not support column metadata"))
```
nalimilan, line 326: Maybe leave this out for now? It's just a convenience function that's easy to replicate by calling `Tables.columnnames` and `colmetadata`, and AFAIK we don't need it currently.

bkamins: We can leave it out. The only difference is performance. If you get `colmetadata(table)` to be `false` you know that NO column metadata is present (so you do not need to check individual columns)

bkamins:
@quinnj - since `colmetadata` is Tables.jl specific do you think it should go to DataAPI.jl or rather to Tables.jl?

src/DataAPI.jl
nalimilan: It seems like we'd better start with the basic API proposed by this PR, and later extend it to cover more complex needs. IIUC in your proposal `metadata(x)` corresponds to self metadata, and `colmetadata(x, col)` to axis metadata (columns are the second axis).

Tokazama: That's essentially correct but I haven't done any bike-shedding with regards to the exact syntax, just the conceptual hierarchy.

Tokazama:
We could give it the same treatment as `eachslice` where there's `eachcol` and `eachrow`.

bkamins:
@pdeffebach - I am having second thoughts here.
Do you feel it will not be too complex if we have both `hasmetadata` and `metadata`. The problem is that with these two functions workflow would be:
1. call `hasmetadata`.
2. If it is `true` (has to be checked with `=== true` check) then call `metadata` to get metadata.

My question is if you feel users would not find it too complex.

The alternative design would be 4 functions:
* `metadata(x, key)` get metadata for object `x` at key `key`; if it is `nothing` then no metadata is defined for this key
* `setmetadata!(x, key, value)` add a mapping for object `x` at key `key` with value `value`;
* `colmetadata(table, col, key)` set metadata for column `col` of table `table` at key `key`; if it is `nothing` then no metadata is defined for this key
* `setcolmetadata!(table, col, key, value)` set metadata for column `col` of table `table` at key `key` with value `value`;

CC @nalimilan

pdeffebach:
Is `nothing` valid metadata? Do we care? I think its probably find. Though I hate doing things like 

```
x = getcolmetadata(df, :c)
if x !== nothing
    println(x)
end
```

to avoid calling `getcolmetadata` twice. Maybe there's a solution to this?

bkamins:
@pdeffebach - you somehow need to check if metadata is present. I do not see any shorter way than what you have written.
Of course the proposed syntax would mean that `nothing` is not allowed as metadata.

As a side note I see we still will need `metadata(x)` and `colmetadata(x)` to return a dictionary to allow for programmatic discovery of keys.

nalimilan:
Having several functions to get or set metadata is indeed a possible alternative to returning a `Dict`. But I'm not sure it's really a win if we need to have `metadata(x)` and `colmetadata(x, col)` to list keys anyway. In the end the risk is that we reproduce the `AbstractDict` API (set, get, get with default, list keys...), just with different names.

Also I don't like returning `nothing` when a key isn't set, as not being able to store `nothing` can be annoying (it creates corner cases; we already have this problem with `findfirst`).

In practice, I imagine that where performance doesn't matter much people will just call `metadata(x)` without first checking `hasmetadata(x)`. And performance likely doesn't matter here, so only (carefully-written) library code will do that check. Anyway, implementations can define a custom lazy `AbstractDict` if they want to avoid allocating a `Dict`. This shouldn't be more work than implementing `setmetadata!(x, key, value)`, `getmetadata(x, key, default)`, `getmetadata(f, x, key)` and so on.

pdeffebach:
@nalimilan Are you saying that you think we shouldn't even have `metadata(df)` *and* `colmetadata(df)`? Cause fwiw I agree, but it looks like there is strong enough movement to explicitely name `colmetadata`

bkamins:
I think @nalimilan wants to keep the API we have agreed on before my proposal to add setters.

Tokazama:
> I think @nalimilan wants to keep the API we have agreed on before my proposal to add setters.

Adding setters brings along a bunch of assumptions with how things are designed per metadata type and use case. It's not impossible to converge on a standard, but it would be nice if we could have something useable here before we begin that potentially long conversation.

nalimilan:
@pdeffebach No I meant that I prefer the current state of the PR. So I'd say everything is ready, though I'd tend to wait until the first implementation in DataFrames is done, to ensure we don't discover yet another tricky issue we didn't anticipate.

test/runtests.jl
```diff
@@ -173,4 +180,21 @@ end
     @test DataAPI.unwrap(missing) === missing
 end
 
+@testset "metadata" begin
+    @test DataAPI.hasmetadata(1) === nothing
+    @test_throws ArgumentError DataAPI.metadata(1)
+    @test DataAPI.hascolmetadata(1, 1) === nothing
+    @test_throws ArgumentError DataAPI.colmetadata(1, 1)
+    @test DataAPI.hascolmetadata(1) === nothing
+    @test_throws ArgumentError DataAPI.colmetadata(1)
```
nalimilan, line 189: This throws a `MethodError` now. Probably not worth testing.
```suggestion
```

bkamins:
OK - so I go ahead with the current state in DataFrames.jl

bkamins:
@nalimilan if table `tab` does not have column `col` then should:
`hasmetadata(tab, col)` return `nothing` or throw an error?

nalimilan:
Good point. I'd say an error to avoid any potential confusion. Though doing that in the fallback isn't super easy. We could define it in Tables.jl instead so that it checks whether the column exists, but AFAIK the only way to do it would be `col in Tables.columnnames(tbl)`, which is O(N) and that cost would be paid even for tables that don't support metadata.

Maybe we can keep the fallback as it is now, but in DataFrames (and other implementations) throw an error?

bkamins:
> Maybe we can keep the fallback as it is now, but in DataFrames (and other implementations) throw an error?

This is what I do now, but I wanted to double check.

nalimilan:
A related point: maybe we should define `hascolmetadata(tbl::Any) = nothing`, so that consumers don't need to check whether metadata is set for each column when the type doesn't support it or the object knows that no metadata has been set for any column?

bkamins:
> A related point: maybe we should define `hascolmetadata(tbl::Any) = nothing`

We have discussed this and decided to drop it for now. I can re-introduce it again. I will finish the https://github.com/JuliaData/DataFrames.jl/pull/3055 PR today, so then you can have a look at it and decide if you think we would find having `hascolmetadata(tbl::Any)` useful. Now #3055 in DataFrames.jl is done without this method.

src/DataAPI.jl
```diff
@@ -287,4 +287,51 @@ using a `sink` function to materialize the table.
...
+"""
+metadata(::T) where {T} =
+    throw(ArgumentError("Objects of type $T do not support metadata"))
+
+"""
+    colmetadata(table, column)
+
+Return metadata associated with column `column` of Tables.jl table `table` as an
+`AbstractDict{String}` object (or an object implementing the same interface).
+
```
bkamins, line 307:
```suggestion
`column` should be any accepted column identifier by `table`, but at least `Symbol` must be supported.
```

src/DataAPI.jl
```diff
@@ -287,4 +287,51 @@ using a `sink` function to materialize the table.
...
+"""
+hasmetadata(::Any) = nothing
+
+"""
+    hascolmetadata(table, column)
+
+Return `true` if column `column` of Tables.jl table `table` has non-empty metadata,
+and return `false` if metadata is empty.
+Return `nothing` if metadata is not supported for `column`.
+
```
bkamins, line 331:
```suggestion


`column` should be any accepted column identifier by `table`, but at least `Symbol` must be supported.

```

bkamins:
As discussed in DataFrames.jl I have added `hascolmetadata(table)` that returns `true` if there is at least one column in a table that has metadata.

bkamins:
@nalimilan - I have updated DataAPI.jl metadata following the discussion on Discourse. In tests I provide a minimal example implementation. Can you please check it?

CC @pdeffebach @nathanrboyer

src/DataAPI.jl
```diff
@@ -287,4 +287,100 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+const STYLE_INFO = """
+One of the uses of the metadata style information is decision
+how the metadata should be propagated when `x` is transformed. This interface
+defines `:none` style that indicates that metadata should not be propagated
```
nalimilan, line 293:
```suggestion
defines the `:none` style that indicates that metadata should not be propagated
```

src/DataAPI.jl
```diff
@@ -287,4 +287,100 @@ using a `sink` function to materialize the table.
...
+If `x` supports metadata, but does not have a mapping for `key` throw `KeyError`.
+
+If `full=true` return a tuple of metadata value and metadata style. Metadata
+style is an additional information about the kind of metadata that is stored
+for the `key`.
+
+$STYLE_INFO
+
+If `default` is passed then return it if `x` does not support metadata or
+does not have a mapping for `key`. The style of `default` value is `:none`.
```
nalimilan, line 312: Shouldn't we also provide  `hasmetadata(x)`, `hasmetadata(x, key)` and `hascolmetadata(x)`, `hascolmetadata(x, col)` and `hascolmetadata(x, col, key)`?

src/DataAPI.jl
```diff
@@ -287,4 +287,100 @@ using a `sink` function to materialize the table.
...
+
+$STYLE_INFO
+
+If `default` is passed then return it if `x` does not support metadata for
+column `col` or does not have a mapping for `key` for column `col`. The style of
+`default` value is `:none`.
+"""
+colmetadata(::T, ::Any, ::Any; full::Bool=false) where {T} =
+    throw(ArgumentError("Objects of type $T do not support getting column metadata"))
+colmetadata(::Any, ::Any, ::Any, default; full::Bool=false) = full ? (default, :none) : default
```
nalimilan, line 358: A potential issue with this fallback is that it won't throw an error if the column doesn't exist. That means that when a type implements metadata, it won't be able to throw in that case either as users may rely on this behavior. I see only two possibilities:
- document this behavior in the docstring (i.e. require implementations to follow it) by adding something like "if `x` does not have column `col`"
- throw an error when type does not support metadata

test/runtests.jl
```diff
@@ -10,6 +10,58 @@ Base.size(x::TestArray) = size(x.x)
...
+end
+
+function DataAPI.colmetadatakeys(x::TestMeta, col)
+    haskey(x.col, col) && return keys(x.col[col])
+    return ()
+end
+
+function DataAPI.colmetadatakeys(x::TestMeta)
+    isempty(x.col) && return ()
+    return Any[col => keys(x.col[col]) for col in keys(x.col)]
```
nalimilan, line 52: Not a big deal for this test, but wouldn't it be better in general to return a generator?
```suggestion
    return (col => keys(x.col[col]) for col in keys(x.col))
```

src/DataAPI.jl
```diff
@@ -287,4 +287,100 @@ using a `sink` function to materialize the table.
...
+const STYLE_INFO = """
+One of the uses of the metadata style information is decision
+how the metadata should be propagated when `x` is transformed. This interface
+defines `:none` style that indicates that metadata should not be propagated
+under transformations. At least this style must be supported by any type
+defining support for metadata.
+"""
+
+"""
+    metadata(x, key, [default]; full::Bool=false)
```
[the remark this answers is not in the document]
bkamins: OK - I will rename `full` to `style`.

> Wouldn't it be better to restrict the type of `key` to strings

This was something I was not sure what is best. The point is that `get` for dictionaries allows passing `key` other than allowed element type of key in the dictionary. I will restrict keys to `AbstractString` and we can assess the consequences.

src/DataAPI.jl
```diff
@@ -287,4 +287,100 @@ using a `sink` function to materialize the table.
...
+and return `x`.
+If `x` does not support setting metadata throw `ArgumentError`.
+
+$STYLE_INFO
+"""
+metadata!(::T, ::Any, ::Any; style) where {T} =
+    throw(ArgumentError("Objects of type $T do not support setting metadata"))
+
+"""
+    colmetadata(x, col, key, [default]; full::Bool=false)
```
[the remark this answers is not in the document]
bkamins: In Tables.jl `col` can be `Symbol` or `Int`. I will change the implementation and docstring accordingly.

src/DataAPI.jl
bkamins: We do not need. if `metadatakeys` returns `()` we know there is no metadata. Otherwise to check if metadata for some `key` is present user should use `key in metadata(x)`. Same with column metadata.

src/DataAPI.jl
bkamins: This is also something I was not sure what to do. I think, given your comment, I will indeed throw an error if the column does not exist.

src/DataAPI.jl
```diff
@@ -287,4 +287,131 @@ using a `sink` function to materialize the table.
...
+If `col` is passed return an iterator of metadata keys for which
+`metadata(x, col, key)` returns a metadata value.
+If `x` does not support metadata for column `col` return `()`.
+
+`col` must be a type that is supported by table `x`. Following the Tables.jl
+contract `Symbol` and `Int` must be supported. However, tables that allow other
+column indexing (e.g., using strings or integers other than `Int`) are allowed
+to accept such types of `col` argument. Passing `col` that is not a column
+of `x` can either throw an error or return `()` (this duality is allowed as
+some Tables.jl do not have schema).
```
bkamins, line 387: @nalimilan - this is the only case left that is problematic. The issue is that we cannot always check if `col` is a column in `x`.

bkamins:
@nalimilan - I have followed your recommendations and pushed new codes. One of the chances is that I have removed the `default` variants as they are not needed in the minimal API and were problematic as it is not always possible to check if `col` is a column of table `x`.

AFAICT the only problematic case left to discuss is https://github.com/JuliaData/DataAPI.jl/pull/48#discussion_r934041357.

Tokazama:
Would it be alright to have `Union{AbstractString,Symbol}` instead of `AbstractString`?

bkamins:
> Would it be alright to have `Union{AbstractString,Symbol}` instead of `AbstractString`?

You mean to allow `Symbol` also as `key` type? The problem is that Arrow.jl, Parquet2.jl, SAS/SPSS/Stata interface would convert this symbol to `String` anyway as there a string is required so after reading the value back we will get a string.
We might alternatively go the DataFrames.jl route, i.e. always store `key` as `Symbol`, but allow passing a request as string and convert it to `Symbol`. This approach would be marginally faster if `Symbol` is used but it is a matter of a few nanoseconds so I am not sure if it is worth it.
@nalimilan - what do you think?

On an implementation note, if we allowed `Symbol` we should define separate methods for it (a minor issue), just like I now define separate methods for `Int` and `Symbol` for `col`. The reason is that `Union{} <: Union{AbstractString,Symbol}` and although impossible in practice this sometimes leads to method dispatch ambiguities and therefore using `Union` should be avoided in interface packages.

Tokazama:
The easiest thing would be to have a dictionary do that conversion for us and this would be a documentation thing. I made a PR to get PropertyDicts.jl up to date (https://github.com/JuliaCollections/PropertyDicts.jl/pull/4), but I don't have any merge rights to move that forward.

bkamins:
> The easiest thing would be to have a dictionary do that conversion for us

The DataAPI.jl interface does not specify an internal implementation for storing metadata. But indeed if we decided to support both strings and Symbols I planned in DataFrames.jl to store keys as `Symbols` and do a conversion of strings to symbols (this is what we do for column names now).

nalimilan:
I don't really see the advantage of allowing symbol keys. Better keep the API simple IMO.

nathanrboyer:
Allowing `AbstractString` and `Symbol` would be more consistent with the rest of DataFrames. I also like with symbols you can use `getproperty` without any surrounding characters like `col.field` if that API style makes it into a convenience package. That said, `String`s are fine if it makes implementation significantly easier.

bkamins:
> That said, Strings are fine if it makes implementation significantly easier.

This is a good point I think. I.e., I think that in low-level API we design here it is better to be more restrictive (as we can always be more flexible in the future). This does not preclude high-level API packages to allow for `Symbol`s (and convert them to strings during the call) if that packages find this useful.

Tokazama:
For the stuff I use it may be a the fields of a type that are the metadata so it's better for me to use a `Symbol`.

Tokazama:
Probably not. I'll just add something that converts strings to symbols. I figured this would be a documentation detail here, since we aren't actually supporting any conversion here.

bkamins:
>  I figured this would be a documentation detail here, since we aren't actually supporting any conversion here.

The point is that this API is intended to ensure inter-package interoperability, so, if I understand @nalimilan correctly, he proposes to keep it minimal. Packages can build on top of these functionalities user-facing API that is more flexible.

pdeffebach:
Tables.jl's `getcolumn` function only allows `Int`s and `Symbol`s. Is that an argument for `Symbol`s only?

bkamins:
>  Is that an argument for `Symbol`s only?

We can discuss to allow `Symbol`s only. But:
1. let us not mix columns and metadata keys - these are two different things.
2. For column access we specify in this PR that `Int` and `Symbol` must be supported in appropriate methods and optionally other column identifiers.
3. For metadata keys - currently we have strings, as this is what Arrow.jl etc. require. Also keys like `"author name"` are more convenient than keys like e.g. `:AuthorName` (that is: I expect that users will use key names that are not valid identifiers). Still If we prefer I am OK to require `Symbol` instead of string and then ask Arrow.jl etc. to convert strings to symbols - this should be lossless.
4. What I do not want is allowing both string and symbols as keys, as it will lead to unnecessary confusion.
5. Note that this is a low-level API that is intended for package interoperability. Specific packages will be allowed to make whatever high-level API they like, and just will be required to perform conversion to string or symbol when low-level API is used.

So the question is - given all these considerations what do we prefer to use strings or symbols?

Tokazama:
In a perfect world we'd just support both, but if it's an issue it might be easier to have strings be the common thing and symbols be supported on an individual basis. It sounds like it will be a lot easier for me to do just do `getmetadata(img::ImageType, key::String) = getmetadata(img, Symbol(key))` than for `DataFrames` to resolve ambiguities do the inverse.

pdeffebach:
Ah sorry, I was confused. 

I agree, we should support only strings for metadata keys.

Tokazama:
I'm not sure if it's within the scope of the current PR, but it would be really nice to have something that acted like `get` so we could do `metadata(data, key, default)`.

src/DataAPI.jl
```diff
@@ -290,97 +290,128 @@ function allcombinations end
...
 One of the uses of the metadata style information is decision
 how the metadata should be propagated when `x` is transformed. This interface
-defines `:none` style that indicates that metadata should not be propagated
+defines the `:none` style that indicates that metadata should not be propagated
 under transformations. At least this style must be supported by any type
 defining support for metadata.
 """
 
+const COL_INFO = """
+`col` must be a type that is supported by table `x`. Following the Tables.jl
```
nalimilan, line 299: Here and in other docstrings, I would rather give information intended at users than intended at package developers, as docstrings are visible to all users. This is mainly a matter of formulation, i.e. say what one can do with the API rather how one is allowed to implement it.

src/DataAPI.jl
nalimilan: Ah right. In theory we could define the fallback only in Tables.jl so that it throws an error when the table has a schema and the column doesn't exist, but maybe this isn't worth the additional complexity. As long as common table types such as DataFrames throw an error in that case, users will likely not rely on the fact that some types accept non-existent columns.

bkamins:
> so we could do `metadata(data, key, default)`

We had it, but I removed it as it is not needed for minimal API. The point is that it can be achieved by:
```
key in metadata(keys) ? metadata(data, key) : default
```
so packages providing higher-level API can add handling of default.

We can add handling of `default` in the future though, after getting some experience with the API. In particular the question is if  `metadata(data, key, default)` should throw an error if `key` is not a string. We can discuss it in the future, but we have already so many open discussions around metadata that I propose to have it in a different PR so that we can move forward faster.

src/DataAPI.jl
bkamins: I have updated the docstrings

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+const STYLE_INFO = """
+One of the uses of the metadata `style` is decision
+how the metadata should be propagated when `x` is transformed. This interface
+defines the `:none` style that indicates that metadata should not be propagated
+under transformations. At least this style is allowed by any type
+defining support for metadata.
```
nalimilan, line 295:
```suggestion
under transformations. All types supporting metadata allow at least this style.
```

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
...
+    metadata(x, key::AbstractString; style::Bool=false)
+
+Return metadata value associated with object `x` for key `key`.
+If `x` does not support metadata throw `ArgumentError`.
+If `x` supports metadata, but does not have a mapping for `key` throw
+`KeyError`.
+
+Passing `key` that is not an `AbstractString` to `metadata` throws
+`MethodError`.
+
```
nalimilan, line 314: I'd drop this as it's to be expected given the signature. Implementations are free to accept more types anyway as long as they document it.
```suggestion
```

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
...
+"""
+    metadata!(x, key::AbstractString, value; style)
+
+Set metadata for object `x` for key `key` to have value `value`
+and style `style` and return `x`.
+If `x` does not support setting metadata throw `ArgumentError`.
+
+Passing `key` that is not an `AbstractString` to `metadata!`
+throws `MethodError`.
+
```
nalimilan, line 341:
```suggestion
```

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
...
+    colmetadata(x, col, key::AbstractString; style::Bool=false)
+
+Return metadata value associated with table `x` for column `col` and key `key`.
+If `x` does not support metadata for column `col` throw `ArgumentError`. If `x`
+supports metadata, but does not have a mapping for column `col` for `key` throw
+`KeyError`.
+
+Passing `key` that is not an `AbstractString` to `colmetadata`
+throws `MethodError`.
+
```
nalimilan, line 357:
```suggestion
```

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
...
+If `col` is passed return an iterator of metadata keys for which `metadata(x,
+col, key)` returns a metadata value. If `x` does not support metadata for column
+`col` return `()`.
+
+`col` must have a type that is supported by table `x` for column indexing.
+Following the Tables.jl contract `Symbol` and `Int` are always allowed. Passing
+`col` that is not a column of `x` can either throws an error (this is a
+preferred behavior if it is possible) or returns `()` (this duality is allowed
+as some Tables.jl tables do not have schema).
```
nalimilan, line 382:
```suggestion
as some Tables.jl tables do not have a schema).
```

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
...
+
+If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
+pairs for all columns that have metadata. If `x` does not support metadata for
+any column return `()`.
```
nalimilan, line 386:
```suggestion
pairs for all columns that have metadata. If `x` does not support column
metadata return `()`.
```

src/DataAPI.jl
```diff
@@ -287,4 +287,125 @@ using a `sink` function to materialize the table.
...
+"""
+    colmetadata!(x, col, key::AbstractString, value; style)
+
+Set metadata for table `x` for column `col` for key `key` to have value `value`
+and style `style`.
+If `x` does not support setting metadata for column `col` throw `ArgumentError`.
+
+Passing `key` that is not an `AbstractString` to `colmetadata!`
+throws `MethodError`.
+
```
nalimilan, line 401:
```suggestion
```

nalimilan (approved):

src/DataAPI.jl
```diff
@@ -287,4 +287,112 @@ using a `sink` function to materialize the table.
...
+"""
+    colmetadatakeys(x, [col])
+
+If `col` is passed return an iterator of metadata keys for which `metadata(x,
+col, key)` returns a metadata value. If `x` does not support metadata for column
+`col` return `()`.
+
+`col` must have a type that is supported by table `x` for column indexing.
+Following the Tables.jl contract `Symbol` and `Int` are always allowed. Passing
+`col` that is not a column of `x` can either throws an error (this is a
```
bkamins, line 370:
```suggestion
`col` that is not a column of `x` can either throw an error (this is a
```

bkamins:
Thank you for the feedback. I think we have settled the design. If there will be no additional comments tomorrow I will start updating https://github.com/JuliaData/DataFrames.jl/pull/3055.

bkamins:
OK - I am starting to implement the API in DataFrames.jl 😄.

bkamins:
I have started working on DataFrames.jl and I already have a decision to be made.
I will focus on table-level metadata as for column level we have the same issue.

In this PR we have `metadata!` for adding metadata. But we will also need a method to drop metadata. There are two issues:
* what would be the best method name for a method that drops metadata for `key` for table `x`. I would propose `deletemetadata!(x, key)`.
* should we define `metadata!` and `deletemetadata!` in DataAPI.jl or it is enough to define accessor methods in DataAPI.jl and leave metadata setting/deleting to be package specific. Thinking of it I would propose to define it in DataAPI.jl, to have shared function names across packages.

bkamins:
@nalimilan - I have added deletion to the API so that we can evaluate if we like it.

nalimilan:
Adding this to DataAPI sounds better than having a separate package, anyway empty definitions are cheap. The `deletemetadata!(x, key)` API sounds good, but I have an hesitation about `deletemetadata!(x)`: shouldn't it be `emptymetadata!(x)` for similarity with the collections API? That's one more function, but the same argument would apply to collections where `delete!(x)` could have been used instead of `empty!(x)`.

bkamins:
I am OK with `emptymetadata!`.

But then for columns we will have:
* `deletemetadata!(x, col, key)`
* `emptymetadata!(x, col)`
* `emptymetadata!(x)`

?

src/DataAPI.jl
```diff
@@ -287,4 +287,158 @@ using a `sink` function to materialize the table.
...
+"""
+metadata!(::T, ::AbstractString, ::Any; style) where {T} =
+    throw(ArgumentError("Objects of type $T do not support setting metadata"))
+
+"""
+    deletemetadata!(x, key::AbstractString)
+
+Delete metadata for object `x` for key `key` and return `x`
+(if metadata for `key` is not present do not perform any action).
+If `key` is not passed delete all metadata for object `x`.
```
bkamins, line 345:
```suggestion
```

src/DataAPI.jl
```diff
@@ -287,4 +287,157 @@ using a `sink` function to materialize the table.
...
+
+`col` must have a type that is supported by table `x` for column indexing.
+Following the Tables.jl contract `Symbol` and `Int` are always allowed. Passing
+`col` that is not a column of `x` either throws an error (this is a
+preferred behavior if it is possible) or returns `()` (this duality is allowed
+as some Tables.jl tables do not have a schema).
+
+If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
+pairs for all columns that have metadata. If `x` does not support column
+metadata return `()`.
```
bkamins, line 395:
```suggestion
If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
pairs for all columns that have metadata, where `col` are `Symbol`.
If `x` does not support column metadata return `()`.
```

src/DataAPI.jl
```diff
@@ -287,4 +287,157 @@ using a `sink` function to materialize the table.
 """
 function allcombinations end
 
+const STYLE_INFO = """
+One of the uses of the metadata `style` is decision
+how the metadata should be propagated when `x` is transformed. This interface
+defines the `:none` style that indicates that metadata should not be propagated
+under transformations. All types supporting metadata allow at least this style.
```
bkamins, line 294: @nalimilan - maybe we should make this description more precise? Currently in DataFrames.jl I needed to make a decision when `:none` metadata is kept and it is kept only in two cases:
* `DataFrame` constructor;
* `copy`;

all other operations drop all `:none` metadata. So, essentially both table level and column level `:none` metadata are attached to a concrete instance of a table or its copies (this is a safest approach, i.e. making sure that indeed when metadata could be invalidated it is dropped). Are we in agreement here?

bkamins: Given no response I will make the definition more precise.

src/DataAPI.jl
```diff
@@ -287,4 +287,158 @@ using a `sink` function to materialize the table.
...
+"""
+    metadatakeys(x)
+
+Return an iterator of metadata keys for which `metadata(x, key)` returns a
+metadata value. If `x` does not support metadata return `()`.
+"""
+metadatakeys(::Any) = ()
+
+"""
+    metadata!(x, key::AbstractString, value; style)
```
nalimilan, line 330: Thinking about it, maybe the syntax would be more natural as `metadata!(x, key => value)`? That would allow extending this in the future to pass multiple pairs if it appears to be convenient.

A counter-argument is that `setindex!` doesn't use that syntax, but it's almost never called that way since `x[key] = value` is nicer. Of course both syntaxes could be allowed as they are not ambiguous (we will probably never allow keys to be pairs).

bkamins: As a first reaction it makes sense.

My only reservation was that in DataFrames.jl `=>` is used for operation specification language, so we would have yet a third way to interpret `=>` there.

The question is how would it look for `colmetadata!`?
```
colmetadata!(x, col, key => value; style=style)
```
(which does not look that nice)

Also note that `metadata!(x, key => value)` would not be allowed, you would need to write `metadata!(x, key => value; style=style)`.

In summary - I would keep the things as they are here and consider what we design here a low-level API.

I assume that the extra new package planned (tentatively named TableMetadataTools.jl) will provide convenient high-level functions. In practice I even expect that if we define there:
```
caption!(table, str) = metadata!(table, "caption", str, style=:note)
caption(table) = metadata(table, "caption")
label!(table, str) = colmetadata!(table, col, "label", str, style=:note)
label(table, col) = colmetadata(table, col, "caption")
```
this will cover 95% of use cases of metadata in practice.

In summary - I propose to discuss a convenience high-level API in TableMetadataTools.jl, as I expect that in that package we will drop the requirement to specify `style` which we have in low-level API, as in high level API all styles will be `:note`.

bkamins:
I changed `:none` to `:default` style that is default without propagation.

@nalimilan - can you please recheck and comment if it can be merged (I guess in https://github.com/JuliaData/DataFrames.jl/pull/3055 we have converged with the implementation). Thank you!

bkamins:
Thank you! The ball starts rolling (the metadata discussion is the most complex addition we made ever in the ecosystem)

Issue #55, opened 2022-11-02
default to metadata! style=:default


nathanrboyer:

bkamins:
@nalimilan - what do you think?

To start with I wanted setting `style` to be explicit, but maybe indeed allowing for `:default` by default is better.

@nathanrboyer - I am working on TableMetadataTools.jl so expect changes in the code.

nathanrboyer:
I can write this at the top of my script to get the desired behavior, but there is probably a cleaner way with metaprogramming:
DataFrames.metadata!(df::DataFrame, key::AbstractString, value::Any) = DataFrames.metadata!(df, key, value; style=:default

nathanrboyer:
If it was named something other than `:default`, then I could see forcing the user to be explicit in the main package and only providing the convenience method in the helper package.

bkamins:
I understand your rationale. Let us wait for @nalimilan to comment. The change will be easy to implement.

nalimilan:
Sounds good.

PR #56, opened 2022-11-03
make it easier to work with metadata by providing more default behaviors

bkamins:
Fixes https://github.com/JuliaData/DataAPI.jl/issues/55
Fixes https://github.com/JuliaData/DataAPI.jl/issues/51

@tokazama - in relation to #51 I ended up recommending to return a dictionary (not just an iterator). Do you think it would be acceptable, or you would prefer to require only an iterator of pairs?

The benefit of dictionary would be that later it is easier for users to work with it. I assume that if a given value does not have already an underlying dictionary to store metadata (most have) then just `Dict` will be used since this is a non-lazy interface.

If someone wants a lazy iterator then using `metadatakeys` and then getting the required metadata is an option that is available.

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -429,7 +429,11 @@ function colmetadata end
 """
     colmetadata(x, [col]; style::Bool=false)
 
-Return a dictionary mapping all column metadata keys to metadata values
+If `col` is not passed return a dictionary mapping columns represented as
+`Symbol` that have associated metadata to dictionaries mapping all
+metadata keys to metadata values associated with table `x` for a given column.
+
+If `col` is passed a dictionary mapping all column metadata keys to metadata values
 associated with column `col` of table `x`. Throw an error if `x` does not
 support reading metadata for column `col` or column `col` is not present in `x`.
 
@@ -438,10 +442,6 @@ style is an additional information about the kind of metadata that is stored for
 the `key`.
 
 $STYLE_INFO
-
-If `col` is not passed return a dictionary mapping columns represented as
-`Symbol` that have associated metadata to dictionaries mapping all
-metadata keys to metadata values associated with table `x` for a given column.
 
 The returned dictionary may be freshly allocated on each call to `colmetadata`
 and is considered to be owned by `x` so it must not be modified.
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -359,7 +359,7 @@ the `key`.
 $STYLE_INFO
 
 The returned dictionary may be freshly allocated on each call to `metadata` and
-is considered to be owned by `x` so it should be only used for reading data.
+is considered to be owned by `x` so it must not be modified.
 """
 function metadata(x::T; style::Bool=false) where {T}
     if !metadatasupport(T).read
@@ -440,12 +440,11 @@ the `key`.
 $STYLE_INFO
 
 If `col` is not passed return a dictionary mapping columns represented as
-`Symbol` that have associated metadata to dictionaries dictionary mapping all
+`Symbol` that have associated metadata to dictionaries mapping all
 metadata keys to metadata values associated with table `x` for a given column.
 
 The returned dictionary may be freshly allocated on each call to `colmetadata`
-and is considered to be owned by `x` so it should be only used for reading
-data.
+and is considered to be owned by `x` so it must not be modified.
 """
 function colmetadata(x::T, col; style::Bool=false) where {T}
     if !colmetadatasupport(T).read
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -330,7 +330,7 @@ The `write` field indicates whether modifying metadata with the [`colmetadata!`]
 colmetadatasupport(::Type) = (read=false, write=false)
 
 """
-    metadata(x, [key::AbstractString], [default]; style::Bool=false)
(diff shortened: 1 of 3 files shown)


Tokazama:
> Do you think it would be acceptable, or you would prefer to require only an iterator of pairs?

I think the tricky part is where that `style` argument comes in. Are we explicitly requiring `valtype(metadata(x)) <: Tuple{Any,Symbol}`?

bkamins:
> Are we explicitly requiring `valtype(metadata(x)) <: Tuple{Any,Symbol}?`

Yes, that is why I assume that this is not lazy. In general this dictionary would have to be freshly allocated (actually both if `style=false` and if `style=true` just with different values). For this reason, if performance matters `metadatakeys` should be used as it is non-allocating

Tokazama:
That makes a lot of sense. Should we have some sort of documentation that makes it clear that `setindex!(metadata(x),...)` isn't the same as `metadata!(x...)`?

src/DataAPI.jl
```diff
@@ -343,6 +343,9 @@ $STYLE_INFO
 
 If `default` is passed then return it if reading metadata is supported but
 mapping for `key` is missing. If `style=true` return `(default, :default)`.
+
+If `key` is not passed return a dictionary mapping all metadata keys to
+metadata values associated with object `x`.
```
nalimilan, line 348: I would document this as a separate method given that a lot of the text above doesn't apply when `key` isn't passed.

nalimilan:
Makes sense. I also think the docs should mention that the resulting object should not be modified as it is owned by `x`, and that it may have to be freshly allocated.

nalimilan:
Also, maybe we should define a fallback `Dict(key => metadata(df, key) for key in metadatakeys(df))`?

src/DataAPI.jl
bkamins: done

bkamins:
Following the comments I have:
1. implemented fallback methods for `metadata` and `colmetadata`
2. split docstings
3. added a comment both for `metadata` and `colmetadata` that the returned dictionary should be only used for reading data and that it may be freshly allocated.

src/DataAPI.jl
```diff
@@ -343,12 +343,31 @@ $STYLE_INFO
...
+with object `x`. Throw an error if `x` does not support reading metadata.
+
+If `style=true` values are tuples of metadata value and metadata style. Metadata
+style is an additional information about the kind of metadata that is stored for
+the `key`.
+
+$STYLE_INFO
+
+The returned dictionary may be freshly allocated on each call to `metadata` and
+is considered to be owned by `x` so it should be only used for reading data.
```
nalimilan, line 362:
```suggestion
is considered to be owned by `x` so it must not be modified.
```

src/DataAPI.jl
```diff
@@ -404,6 +426,42 @@ If `style=true` return `(default, :default)`.
...
+support reading metadata for column `col` or column `col` is not present in `x`.
+
+If `style=true` values are tuples of metadata value and metadata style. Metadata
+style is an additional information about the kind of metadata that is stored for
+the `key`.
+
+$STYLE_INFO
+
+If `col` is not passed return a dictionary mapping columns represented as
+`Symbol` that have associated metadata to dictionaries dictionary mapping all
```
nalimilan, line 443:
```suggestion
`Symbol` that have associated metadata to dictionaries mapping all
```

src/DataAPI.jl
```diff
@@ -404,6 +426,42 @@ If `style=true` return `(default, :default)`.
...
+metadata keys to metadata values associated with table `x` for a given column.
+
+The returned dictionary may be freshly allocated on each call to `colmetadata`
+and is considered to be owned by `x` so it should be only used for reading
+data.
```
nalimilan, line 448:
```suggestion
and is considered to be owned by `x` so it must not be modified.
```

src/DataAPI.jl
nalimilan, line 443: Maybe put this in the first paragraph, as it completely changes the behavior of the function? (It could amost make sense to have a separate docstring but that would be quite heavy...)

bkamins: OK - done.

bkamins:
@ExpandingMan - can you please have a look at this PR before it is merged and released (it will have a minor in fluence on Parquet2.jl implementation, but actually it will go along what you partially have already). Thank you!

src/DataAPI.jl
```diff
@@ -429,7 +429,11 @@ function colmetadata end
 """
     colmetadata(x, [col]; style::Bool=false)
 
-Return a dictionary mapping all column metadata keys to metadata values
+If `col` is not passed return a dictionary mapping columns represented as
+`Symbol` that have associated metadata to dictionaries mapping all
+metadata keys to metadata values associated with table `x` for a given column.
+
+If `col` is passed a dictionary mapping all column metadata keys to metadata values
```
nalimilan, line 436:
```suggestion
If `col` is passed return a dictionary mapping all column metadata keys to metadata values
```

ExpandingMan:
Looks good to me.

I haven't actually tested anything against this but it looks non-breaking.

nalimilan (approved):

bkamins:
Thank you!

PR #58, opened 2022-12-04
Add operator kwarg to Cols

bkamins:
x-ref https://github.com/JuliaData/DataFrames.jl/pull/3224

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -198,26 +198,24 @@ struct All{T<:Tuple}
 end
 
 """
-    Cols(cols...; operation::Symbol=:union)
-    Cols(f::Function; operation::Symbol=:union)
+    Cols(cols...; operation=union)
+    Cols(f::Function; operation=union)
 
 Select columns matching specifications in `cols`. If `cols == ()`, select no columns.
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
 
-When multiple `cols` selectors are passed,  if `operation=:union` (the default)
-all columns matching at least one selector are returned, and if
-`operation=:intersect` only columns matching all selectors are returned.
+When multiple `cols` selectors are passed then the sets of columns selected by them
+are passed to `operation` callable as positional arguments.
+`operation` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
+defined in Base Julia. By default `operation=union` in which case all columns matching
+at least one selector are returned.
 """
 struct Cols{T<:Tuple}
     cols::T
-    operation::Symbol
-    function Cols(args...; operation::Symbol=:union)
-        if !(operation == :union || operation == :intersect)
-            throw(ArgumentError("operation must be `:union` or `:intersect`"))
-        end
-        return new{typeof(args)}(args, operation)
+    operation
+    Cols(args...; operation::Symbol=union) = new{typeof(args)}(args, operation)
     end
 end
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -183,9 +183,9 @@ Between(x::Union{Int, Symbol}, y::AbstractString) = Between(x, Symbol(y))
 Between(x::AbstractString, y::Union{Int, Symbol}) = Between(Symbol(x), y)
 
 """
-    All(cols...)
+    All()
 
-Select the union of the selections in `cols`. If `cols == ()`, select all columns.
+Select all columns.
 """
 struct All{T<:Tuple}
     cols::T
@@ -198,17 +198,27 @@ struct All{T<:Tuple}
 end
 
 """
-    Cols(cols...)
-    Cols(f::Function)
+    Cols(cols...; operation::Symbol=:union)
+    Cols(f::Function; operation::Symbol=:union)
 
-Select the union of the selections in `cols`. If `cols == ()`, select no columns.
+Select a union of the selections in `cols`. If `cols == ()`, select no columns.
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
+
+If `operation` keyword argument is passed then it can be either `:union` (the default)
+or `:intersect`. If it is `:intersect` then an intersection of the selections
+in `cols` is selected.
 """
 struct Cols{T<:Tuple}
     cols::T
-    Cols(args...) = new{typeof(args)}(args)
+    operation::Symbol
+    function Cols(args...; operation::Symbol=:union)
+        if !(operation == :union || operation == :intersect)
+            throw(ArgumentError("operation must be `:union` or `:intersect`"))
+        end
+        return new{typeof(args)}(args, operation)
+    end
 end
 
 """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -216,7 +216,6 @@ struct Cols{T<:Tuple}
     cols::T
     operation
     Cols(args...; operation::Symbol=union) = new{typeof(args)}(args, operation)
-    end
 end
 
 """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -215,7 +215,7 @@ at least one selector are returned.
 struct Cols{T<:Tuple}
     cols::T
     operation
-    Cols(args...; operation::Symbol=union) = new{typeof(args)}(args, operation)
+    Cols(args...; operation=union) = new{typeof(args)}(args, operation)
 end
 
 """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -198,8 +198,8 @@ struct All{T<:Tuple}
 end
 
 """
-    Cols(cols...; operation=union)
-    Cols(f::Function; operation=union)
+    Cols(cols...; operator=union)
+    Cols(f::Function; operator=union)
 
 Select columns matching specifications in `cols`. If `cols == ()`, select no columns.
 
@@ -207,15 +207,15 @@ If the only positional argument is a `Function` `f` then select the columns whos
 names passed to the `f` predicate as strings return `true`.
 
 When multiple `cols` selectors are passed then the sets of columns selected by them
-are passed to `operation` callable as positional arguments.
-`operation` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
-defined in Base Julia. By default `operation=union` in which case all columns matching
+are passed to `operator` callable as positional arguments.
+`operator` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
+defined in Base Julia. By default `operator=union` in which case all columns matching
 at least one selector are returned.
 """
 struct Cols{T<:Tuple}
     cols::T
-    operation
-    Cols(args...; operation=union) = new{typeof(args)}(args, operation)
+    operator
+    Cols(args...; operator=union) = new{typeof(args)}(args, operator)
 end
 
 """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -201,14 +201,14 @@ end
     Cols(cols...; operation::Symbol=:union)
     Cols(f::Function; operation::Symbol=:union)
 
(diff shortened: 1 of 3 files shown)


bkamins:
CC @jar @pdeffebach @yjunechoe

src/DataAPI.jl
```diff
@@ -183,9 +183,9 @@ Between(x::Union{Int, Symbol}, y::AbstractString) = Between(x, Symbol(y))
 Between(x::AbstractString, y::Union{Int, Symbol}) = Between(Symbol(x), y)
 
 """
```
bkamins, line 185: this change removes long deprecated old behavior description.

src/DataAPI.jl
```diff
@@ -183,9 +183,9 @@ Between(x::Union{Int, Symbol}, y::AbstractString) = Between(x, Symbol(y))
...
-    All(cols...)
+    All()
 
-Select the union of the selections in `cols`. If `cols == ()`, select all columns.
+Select all columns.
```
bkamins, line 188: this change removes long deprecated old behavior description.

src/DataAPI.jl
```diff
@@ -198,17 +198,27 @@ struct All{T<:Tuple}
 end
 
 """
-    Cols(cols...)
-    Cols(f::Function)
+    Cols(cols...; operation::Symbol=:union)
+    Cols(f::Function; operation::Symbol=:union)
 
-Select the union of the selections in `cols`. If `cols == ()`, select no columns.
+Select a union of the selections in `cols`. If `cols == ()`, select no columns.
```
nalimilan, line 204:
```suggestion
Select columns matching specifications in `cols`. If `cols == ()`, select no columns.
```

src/DataAPI.jl
```diff
@@ -198,17 +198,27 @@ struct All{T<:Tuple}
...
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
+
+If `operation` keyword argument is passed then it can be either `:union` (the default)
+or `:intersect`. If it is `:intersect` then an intersection of the selections
+in `cols` is selected.
```
nalimilan, line 211:
```suggestion
When multiple `cols` selectors are passed,  if `operation=:union` (the default)
all columns matching at least one selector are returned, and if
`operation=:intersect` only columns matching all selectors are returned.
```

nalimilan (approved):

adienes:
I think we discussed this in the previous issue but I forgot---could you remind me why we use keywords here rather than a function? Then it would have `Cols(...; operation=any)` and `Cols(...; operation=all)` using built in functions `any` and `all`, and could use user-defined predicates.

bkamins:
We have not discussed it previously. Actually what you propose is a good idea. But `any` and `all` are not correct functions to be passed. We can say that `union` function is the default and `intersect` is a function to pass if one wants intersection. However, potentially user could pass any function operating on sets and returning a set as a result.

Then the definition in DataFrames.jl would be:
```
@inline Base.getindex(x::AbstractIndex, idx::Cols) =
    isempty(idx.cols) ? Int[] : idx.operation(getindex.(Ref(x), idx.cols)...)
```
@nalimilan - what do you think?

adienes:
Makes sense. Then we'd also get `setdiff` and `symdiff` for free

jariji:
Yep, I agree with adienes.

nalimilan:
Yeah why not. Maybe we need a more specific name than "operation" then for the argument. Again the same discussion as the one about "combine" in `unstack`? :-)

bkamins:
AFAICT this fits well. In mathematics `union` and `intersect` etc. are typically called "operations", c.f. e.g. https://en.wikipedia.org/wiki/Set_(mathematics)#Basic_operations

src/DataAPI.jl
```diff
@@ -198,17 +198,27 @@ struct All{T<:Tuple}
 end
 
 """
-    Cols(cols...)
-    Cols(f::Function)
+    Cols(cols...; operation::Symbol=:union)
+    Cols(f::Function; operation::Symbol=:union)
```
bkamins, line 202:
```suggestion
    Cols(cols...; operation=union)
    Cols(f::Function; operation=union)
```

src/DataAPI.jl
```diff
@@ -198,17 +198,27 @@ struct All{T<:Tuple}
...
 
-Select the union of the selections in `cols`. If `cols == ()`, select no columns.
+Select columns matching specifications in `cols`. If `cols == ()`, select no columns.
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
+
+When multiple `cols` selectors are passed,  if `operation=:union` (the default)
+all columns matching at least one selector are returned, and if
+`operation=:intersect` only columns matching all selectors are returned.
```
bkamins, line 211:
```suggestion
When multiple `cols` selectors are passed then the sets of columns selected by them
are passed to `operation` callable as positional arguments.
`operation` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
defined in Base Julia. By default `operation=union` in which case all columns matching
at least one selector are returned.
```

src/DataAPI.jl
```diff
@@ -198,17 +198,27 @@ struct All{T<:Tuple}
...
 struct Cols{T<:Tuple}
     cols::T
-    Cols(args...) = new{typeof(args)}(args)
+    operation::Symbol
+    function Cols(args...; operation::Symbol=:union)
+        if !(operation == :union || operation == :intersect)
+            throw(ArgumentError("operation must be `:union` or `:intersect`"))
+        end
+        return new{typeof(args)}(args, operation)
+    end
```
bkamins, line 221:
```suggestion
    operation
    Cols(args...; operation::Symbol=union) = new{typeof(args)}(args, operation)
    end
```

test/runtests.jl
```diff
@@ -249,6 +249,9 @@ end
             (Ref(DataAPI.BroadcastedSelector(v())) .=> [sum, float])
     end
 
+    @test DataAPI.Cols(:a, operation=:union).operation == :union
+    @test DataAPI.Cols(:a, operation=:intersect).operation == :intersect
+    @test_throws ArgumentError DataAPI.Cols(:a, operation=:bad)
```
bkamins, line 254:
```suggestion
    @test DataAPI.Cols(:a, operation=union).operation == union
    @test DataAPI.Cols(:a, operation=intersect).operation == intersect
```

src/DataAPI.jl
```diff
@@ -198,17 +198,25 @@ struct All{T<:Tuple}
...
+`operation` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
+defined in Base Julia. By default `operation=union` in which case all columns matching
+at least one selector are returned.
 """
 struct Cols{T<:Tuple}
     cols::T
-    Cols(args...) = new{typeof(args)}(args)
+    operation
+    Cols(args...; operation::Symbol=union) = new{typeof(args)}(args, operation)
+    end
```
bkamins, line 219:
```suggestion
```

src/DataAPI.jl
```diff
@@ -198,17 +198,24 @@ struct All{T<:Tuple}
...
+are passed to `operation` callable as positional arguments.
+`operation` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
+defined in Base Julia. By default `operation=union` in which case all columns matching
+at least one selector are returned.
 """
 struct Cols{T<:Tuple}
     cols::T
-    Cols(args...) = new{typeof(args)}(args)
+    operation
+    Cols(args...; operation::Symbol=union) = new{typeof(args)}(args, operation)
```
bkamins, line 218:
```suggestion
    Cols(args...; operation=union) = new{typeof(args)}(args, operation)
```

bkamins:
I have updated the PR.

jariji:
The tests don't actually test the functionality.

adienes:
Perhaps `operator` is very slightly a better choice for some very tenuous reasons:

1. often the arguments in `Cols` will be indicator functions of sets, and a function that acts on other functions is often called an operator
2. to me it scans slightly more clearly that what follows `operator=` should be a verb (i.e. a function)
3. it is 1 character shorter :)

The wikipedia page for the [Algebra of Sets](https://en.wikipedia.org/wiki/Algebra_of_sets) uses both terms interchangeably, so I'm not sure which is more standard.

bkamins:
I checked several references, and the border between `operation` and `operator` is thin. [Operation (mathematics)](https://en.wikipedia.org/wiki/Operation_(mathematics)) gives a most clear distinction:

> An operator is similar to an operation in that it refers to the symbol or the process used to denote the operation, hence their point of view is different. For instance, one often speaks of "the operation of addition" or "the addition operation", when focusing on the operands and result, but one switches to "addition operator" (rarely "operator of addition"), when focusing on the process

Given this explanation indeed `operator` seems better, but I am not sure.

src/DataAPI.jl
```diff
@@ -198,17 +198,24 @@ struct All{T<:Tuple}
 end
 
 """
-    Cols(cols...)
-    Cols(f::Function)
+    Cols(cols...; operation=union)
+    Cols(f::Function; operation=union)
```
bkamins, line 202:
```suggestion
    Cols(cols...; operator=union)
    Cols(f::Function; operator=union)
```

src/DataAPI.jl
```diff
@@ -198,17 +198,24 @@ struct All{T<:Tuple}
...
-Select the union of the selections in `cols`. If `cols == ()`, select no columns.
+Select columns matching specifications in `cols`. If `cols == ()`, select no columns.
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
+
+When multiple `cols` selectors are passed then the sets of columns selected by them
+are passed to `operation` callable as positional arguments.
+`operation` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
+defined in Base Julia. By default `operation=union` in which case all columns matching
```
bkamins, line 212:
```suggestion
are passed to `operator` callable as positional arguments.
`operator` should be a set operation function, like `union`, `intersect`, `setdiff`, and `symdiff`
defined in Base Julia. By default `operator=union` in which case all columns matching
```

src/DataAPI.jl
```diff
@@ -198,17 +198,24 @@ struct All{T<:Tuple}
...
+at least one selector are returned.
 """
 struct Cols{T<:Tuple}
     cols::T
-    Cols(args...) = new{typeof(args)}(args)
+    operation
+    Cols(args...; operation=union) = new{typeof(args)}(args, operation)
```
bkamins, line 218:
```suggestion
    operator
    Cols(args...; operator=union) = new{typeof(args)}(args, operator)
```

test/runtests.jl
```diff
@@ -249,6 +249,8 @@ end
             (Ref(DataAPI.BroadcastedSelector(v())) .=> [sum, float])
     end
 
+    @test DataAPI.Cols(:a, operation=union).operation == union
+    @test DataAPI.Cols(:a, operation=intersect).operation == intersect
```
bkamins, line 253:
```suggestion
    @test DataAPI.Cols(:a, operator=union).operator == union
    @test DataAPI.Cols(:a, operator=intersect).operator == intersect
```

src/DataAPI.jl
```diff
@@ -198,17 +198,24 @@ struct All{T<:Tuple}
...
+    Cols(f::Function; operator=union)
 
-Select the union of the selections in `cols`. If `cols == ()`, select no columns.
+Select columns matching specifications in `cols`. If `cols == ()`, select no columns.
 
 If the only positional argument is a `Function` `f` then select the columns whose
 names passed to the `f` predicate as strings return `true`.
+
+When multiple `cols` selectors are passed then the sets of columns selected by them
+are passed to `operator` callable as positional arguments.
```
nalimilan, line 210:
```suggestion
are passed to `operator` as positional arguments.
```
(or "the `operator` callable", but the next sentence says it must be a function)

nalimilan (approved):

bkamins:
Thank you!

PR #62, opened 2023-07-07
Suppor dimensional metadata

Tokazama:
The core metadata methods (`metadata`, `metadata!`, etc.) provide an intuitive method of communicating that a key-value pair provides some information about the instance of the object it is attached to. The column metadata methods similarly provide an intuitive method of communicating the a key-value pair provides some information about the column of a particular object. However, data with multiple dimensions (array, mesh, coordinates, etc.) often have metadata stored per dimension.

This PR provides support for metadata corresponding to specific dimensions. New method implemented here follow the name pattern of column metadata methods (`colmetadata` to `dimmetadata`).

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -622,7 +622,7 @@ function deletedimmetadata! end
 """
     emptydimmetadata!(x, dim::Int)
 
-Delete all metadata for `x` corresponding to dimension `dim` and return `x`.
+Delete all metadata for object `x` for to dimension `dim` and return `x`.
 Throw an error if `x` does not support metadata deletion for dimension `dim`.
 """
 function emptydimmetadata! end
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -602,7 +602,7 @@ function dimmetadatakeys end
 """
     dimmetadata!(x, dim::Int, key::AbstractString, value; style::Symbol=:default)
 
-Set metadata for `x` for dimension `dim` for key `key` to have value `value`
+Set metadata for object `x` for dimension `dim` for key `key` to have value `value`
 and style `style` (`:default` by default) and return `x`.
 Throw an error if `x` does not support setting metadata for dimension `dim`.
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -573,7 +573,7 @@ function dimmetadata(x::T, dim::Int; style::Bool=false) where {T}
     end
     return Dict(key => dimmetadata(x, dim, key, style=style) for key in dimmetadatakeys(x, dim))
 end
-function dimmetadata(x::T; style::Bool=false) where {T}
+function dimmetadata(x; style::Bool=false)
     Tuple(dimmetadata(x, dim; style=style) for dim in 1:ndims(x))
 end
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -623,6 +623,7 @@ function deletedimmetadata! end
     emptydimmetadata!(x, dim::Int)
 
 Delete all metadata for object `x` for to dimension `dim` and return `x`.
+If `dim` is not passed delete all dimension level metadata for table `x`.
 Throw an error if `x` does not support metadata deletion for dimension `dim`.
 """
 function emptydimmetadata! end
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -562,7 +562,7 @@ the `key`.
 
 $STYLE_INFO
 
-The returned dictionary may be freshly allocated on each call to `dimmetadata` and
+The returned object may be freshly allocated on each call to `dimmetadata` and
 is considered to be owned by `x` so it must not be modified.
 """
 function dimmetadata(x::T, dim::Int; style::Bool=false) where {T}
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -381,6 +381,9 @@ end
 Return an iterator of metadata keys for which `metadata(x, key)` returns a
 metadata value.
 Throw an error if `x` does not support reading metadata.
+
+If `metadatasupport(typeof(x)).read` or `metadatasupport(typeof(x)).write` return `true`
+this should also be defined.
 """
 function metadatakeys end
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -562,7 +562,8 @@ values associated with object `x` for a given dimension, so that
 If `dim` is passed return a dictionary mapping all column metadata keys
 to metadata values associated with dimension `dim` of object `x`, so that
 `colmetadata(x, dim)[key] == dimmetadata(x, dim, key)`.
-Throw an error if `x` does not support reading metadata corresponding to dimension `dim`.
+Throw an error if `x` does not support reading metadata for dimension `dim`
+or `dim` is not dimension of `x`.
 
 If `style=true` values are tuples of metadata value and metadata style. Metadata
 style is an additional information about the kind of metadata that is stored for
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -571,9 +571,14 @@ function dimmetadata(x::T, dim::Int; style::Bool=false) where {T}
 end
 
 """
-    dimmetadatakeys(x, dim::Int)
+    dimmetadatakeys(x, [dim::Int])
 
-Return an iterator of keys for corresponding to metadata for dimension `dim` of `x`.
+If `dim` is passed return an iterator of metadata keys for which
+`metadata(x, dim, key)` returns a metadata value. Throw an error if `x` does not
+support reading column metadata or if `dim` is not a dimension of `x`.
+
+If `dim` is not passed return an iterator of `dim => colmetadatakeys(x, dim)`
+pairs for all dimensions that have metadata.
 If `x` does not support dimension metadata return `()`.
 """
 function dimmetadatakeys end
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -554,9 +554,14 @@ function dimmetadata end
 """
     dimmetadata(x, [dim::Int]; style::Bool=false)
 
-Return a dictionary of metadata corresponding to dimension `dim` of object `x`. If `dim`
- is not passed return a collection mapping each dimension to its metadata so that
-`collection[dim][key] == dimmetadata(x, dim, key)`.
+If `dim` is not passed, return a dictionary mapping dimensions that have
+associated metadata to dictionaries mapping all metadata keys to metadata
+values associated with object `x` for a given dimension, so that
+`colmetadata(x)[dim][key] == dimmetadata(x, dim, key)`.
+
+If `dim` is passed return a dictionary mapping all column metadata keys
+to metadata values associated with dimension `dim` of object `x`, so that
+`colmetadata(x, dim)[key] == dimmetadata(x, dim, key)`.
 Throw an error if `x` does not support reading metadata corresponding to dimension `dim`.
 
 If `style=true` values are tuples of metadata value and metadata style. Metadata
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -480,6 +480,7 @@ Following the Tables.jl contract `Symbol` and `Int` are always allowed.
 
 If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
 pairs for all columns that have metadata, where `col` are `Symbol`.
+If `x` does not support column metadata return `()`.
 
 If `colmetadatasupport(typeof(x)).read` or `colmetadatasupport(typeof(x)).write` return `true`
 this should also be defined.
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -323,20 +323,6 @@ The `write` field indicates whether modifying metadata with the [`metadata!`](@r
 metadatasupport(::Type) = (read=false, write=false)
 
 """
-    colmetadatasupport(T::Type)
-
-Return a `NamedTuple{(:read, :write), Tuple{Bool, Bool}}` indicating whether
-values of type `T` support column metadata.
-
-The `read` field indicates whether reading metadata with the [`colmetadata`](@ref)
-and [`colmetadatakeys`](@ref) functions is supported.
-
-The `write` field indicates whether modifying metadata with the [`colmetadata!`](@ref),
-[`deletecolmetadata!`](@ref), and [`emptycolmetadata!`](@ref) functions is supported.
-"""
-colmetadatasupport(::Type) = (read=false, write=false)
-
-"""
     metadata(x, key::AbstractString, [default]; style::Bool=false)
 
 Return metadata value associated with object `x` for key `key`. Throw an error
@@ -416,6 +402,20 @@ Throw an error if `x` does not support metadata deletion.
 function emptymetadata! end
 
 """
+    colmetadatasupport(T::Type)
+
+Return a `NamedTuple{(:read, :write), Tuple{Bool, Bool}}` indicating whether
+values of type `T` support column metadata.
+
+The `read` field indicates whether reading metadata with the [`colmetadata`](@ref)
+and [`colmetadatakeys`](@ref) functions is supported.
+
+The `write` field indicates whether modifying metadata with the [`colmetadata!`](@ref),
+[`deletecolmetadata!`](@ref), and [`emptycolmetadata!`](@ref) functions is supported.
+"""
+colmetadatasupport(::Type) = (read=false, write=false)
+
+"""
     colmetadata(x, col, key::AbstractString, [default]; style::Bool=false)
 
 Return metadata value associated with table `x` for column `col` and key `key`.
@@ -485,8 +485,8 @@ If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
 pairs for all columns that have metadata, where `col` are `Symbol`.
 If `x` does not support column metadata return `()`.
 
-This method must be defined if `dimmetadatasupport(typeof(x)).read` or
-`dimmetadatasupport(typeof(x)).write` return `true`.
+This method must be defined if `colmetadatasupport(typeof(x)).read` or
+`colmetadatasupport(typeof(x)).write` return `true`.
 """
 function colmetadatakeys end
 
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -516,6 +516,97 @@ Throw an error if `x` does not support metadata deletion for column `col`.
 function emptycolmetadata! end
 
 """
+    dimmetadatasupport(T::Type, dim::Int)
+
+Return a `NamedTuple{(:read, :write), Tuple{Bool, Bool}}` indicating whether
+values of type `T` support metadata correspond to dimension `dim`.
+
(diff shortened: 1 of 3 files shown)


quinnj:
This seems fine to me? @bkamins any concerns?

test/runtests.jl
```diff
@@ -17,12 +17,22 @@ DataAPI.levels(x::TestArray) = reverse(DataAPI.levels(x.x))
...
-    TestMeta() = new(Dict{String, Any}(), Dict{Symbol, Dict{String, Any}}())
+    function TestMeta()
+        new(Dict{String, Any}(), Dict{Symbol, Dict{String, Any}}(), (Dict{String, Any}(), Dict{String, Any}()))
+    end
 end
 
 DataAPI.metadatasupport(::Type{TestMeta}) = (read=true, write=true)
 DataAPI.colmetadatasupport(::Type{TestMeta}) = (read=true, write=true)
+function DataAPI.dimmetadatasupport(::Type{TestMeta}, dim::Int)
+    if dim === 1 || dim === 2
```
bkamins, line 30:
```suggestion
    if dim == 1 || dim == 2
```

test/runtests.jl
```diff
@@ -301,12 +343,25 @@ end
     @test_throws MethodError DataAPI.colmetadatakeys(1, 1)
     @test_throws MethodError DataAPI.colmetadatakeys(1)
 
+    dim = 1
```
bkamins, line 346: line not needed?

test/runtests.jl
```diff
@@ -353,6 +408,26 @@ end
     @test DataAPI.colmetadata!(tm, :col, "a", "100", style=:note) == tm
     DataAPI.emptycolmetadata!(tm)
     @test isempty(DataAPI.colmetadatakeys(tm))
+
+    dim = 1
```
bkamins, line 412: same here

bkamins:
I think it is OK to add. I assume @Tokazama has some concrete use case (i.e. I would avoid adding functionalities that will not be used in practice). I have went through the code and left some small comments.

Tokazama:
> I think it is OK to add. I assume @Tokazama has some concrete use case (i.e. I would avoid adding functionalities that will not be used in practice). I have went through the code and left some small comments.

Besides the use cases in the original post I have images that have metadata per dimension in the form of name, per dim machine acquisition information, and additional metadata along axes (similar to colmetadata).

test/runtests.jl
bkamins: Ah - now I see how you use it. But why do you set it twice then and do I understand it correctly that it is always `1`?

bkamins (approved):

bkamins:
@quinnj - all looks good. OK to merge? (I would make a release next)

Tokazama:
Anything needed on my end?

bkamins:
Needs a second approval (which might require some changes, but most likely not) before being merged and released.

bkamins:
@nalimilan - do you have an opinion on this PR?

src/DataAPI.jl
```diff
@@ -515,6 +515,97 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+
+The `write` field indicates whether modifying metadata with the [`dimmetadata!`](@ref),
+[`deletemetadata!`](@ref), and [`emptymetadata!`](@ref) functions is supported.
+"""
+dimmetadatasupport(::Type, i::Int) = (read=false, write=false)
+
+"""
+    dimmetadata(x, dim::Int, key::AbstractString, [default]; style::Bool=false)
+
+Return metadata value associated with `x` at dimension `dim` and key `key`.
```
nalimilan, line 535: I'd say "for dimension `dim`" rather than "at" for consistency with `colmetadata` (here and elsewhere).

src/DataAPI.jl
```diff
@@ -515,6 +515,97 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+Return an iterator of keys for corresponding to metadata for dimension `dim` of `x`.
+If `x` does not support dimension metadata return `()`.
+"""
+function dimmetadatakeys end
+
+"""
+    dimmetadata!(x, dim::Int, key::AbstractString, value; style::Symbol=:default)
+
+Set the metadata value and style associated with `x` at dimension `dim` for key `key` to `value`
+and `style`, respectively. (`:default` by default) and return `x`.
```
nalimilan, line 585: Misplaced dot. Better reuse exactly the same wording as for `colmetadata!` for consistency.

test/runtests.jl
nalimilan: Why not write `1` instead of using a variable? That sounds clearer to me.

nalimilan (commented):
Makes sense.

Tokazama:
> Why not write 1 instead of using a variable? That sounds clearer to me.

The column metadata tests use `:col` as the argument for which column is being accessed. I thought using the variable `dim` instead of just `1` would make it easier to quickly read and understand what exactly was being tested.

test/runtests.jl
nalimilan: > The column metadata tests use `:col` as the argument for which column is being accessed. I thought using the variable `dim` instead of just `1` would make it easier to quickly read and understand what exactly was being tested.

@Tokazama I'm not convinced using `dim` is clearer, as you have to check what `dim` corresponds to. Also as you note the other tests using `:col` repeat it for each line so that's not consistent.

Tokazama: If you think replacing it with '1' will improve readability I'm happy to change it

nalimilan: Please do.

src/DataAPI.jl
```diff
@@ -515,6 +515,97 @@ Throw an error if `x` does not support metadata deletion for column `col`.
 """
 function emptycolmetadata! end
 
+"""
+    dimmetadatasupport(T::Type, dim::Int)
+
+Return a `NamedTuple{(:read, :write), Tuple{Bool, Bool}}` indicating whether
+values of type `T` support metadata correspond to dimension `dim`.
```
nalimilan, line 522: Sorry, I missed this substantial point: `colmetadatasupport` takes a single argument, i.e. it asks whether column metadata is supported at all, without considering a particular column. Shouldn't we do the same here? Once a type supports dimension metadata, it can return `nothing` for dimensions that do not have any metadata.

src/DataAPI.jl
```diff
@@ -515,6 +515,97 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+        throw(ArgumentError("Objects of type $T do not support reading dimension metadata"))
+    end
+    return Dict(key => dimmetadata(x, dim, key, style=style) for key in dimmetadatakeys(x, dim))
+end
+
+"""
+    dimmetadatakeys(x, dim::Int)
+
+Return an iterator of keys for corresponding to metadata for dimension `dim` of `x`.
+If `x` does not support dimension metadata return `()`.
```
nalimilan, line 577: For consistency with `colmetadatakeys`, the behavior should be something like this:
```suggestion
    dimmetadatakeys(x, [dim::Int])

If `dim` is passed return an iterator of metadata keys for which
`metadata(x, dim, key)` returns a metadata value. Throw an error if `x` does not
support reading column metadata or if `dim` is not a dimension of `x`.

If `dim` is not passed return an iterator of `dim => colmetadatakeys(x, dim)`
pairs for all dimensions that have metadata.
If `x` does not support dimension metadata return `()`.
```

src/DataAPI.jl
```diff
@@ -515,6 +515,97 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+$STYLE_INFO
+
+If `default` is passed then return it if `x` supports reading metadata and has
+dimension `dim` but mapping for `key` is missing.
+If `style=true` return `(default, :default)`.
+"""
+function dimmetadata end
+
+"""
+    dimmetadata(x, dim::Int; style::Bool=false)
```
nalimilan, line 552: How about providing a single-argument function here too?

bkamins:
> I wonder how this interacts with the existence of column metadata, as columns are usually considered as dimension 2.

I assumed this is completely detached. I.e. `colmetadata` is for objects that implement Tables.jl interface. And this PR is for array-like objects. Note that array-like objects can be re-interpreted as tables in many different ways.

src/DataAPI.jl
nalimilan: What do you mean by "there is exactly one dimension always"? :-)

src/DataAPI.jl
nalimilan: Yeah, I would do exactly the same as for `colmetadata(x)`:
```julia
If `col` is not passed return a dictionary mapping dimensions
that have associated metadata to dictionaries mapping all
metadata keys to metadata values associated with table `x` for a given dimension.
```

Tokazama: Oh, that's a good idea!

Tokazama:
> > I assumed this is completely detached. I.e. `colmetadata` is for objects that implement Tables.jl interface. And this PR is for array-like objects. Note that array-like objects can be re-interpreted as tables in many different ways.
> 
> @bkamins Sure. But e.g. for a `DataFrame` we could imagine implementing `dimmetadata` at some point given that we define `ndims`, `axes`, `getindex`... So I wanted to make sure we consider whether that would make sense.
> 
> > The way to support metadata per element along a dimension with this would be a specific "indices" or "axis" key whose value can be used somewhat like a tables column metadata.
> 
> @Tokazama I guess that would work. Then you mean that the intended use case for this PR is both metadata about a dimension itself and metadata about its elements (rows, columns...), right?

Exactly! I figured the alternative was having an additional metadata set of methods (e.g., `axesmetadata`) which seemed a bit over the top since one can still accomplish similar functionality with the dim methods here. The overlap with column metadata could also be seen as redundant but tables have consistently proven to be very unique since each column can be conceptualized as its own entity.

src/DataAPI.jl
```diff
@@ -515,6 +515,113 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+
+Return a dictionary of metadata corresponding to dimension `dim` of object `x`. If `dim`
+ is not passed return a collection mapping each dimension to its metadata so that
+`collection[dim][key] == dimmetadata(x, dim, key)`.
+Throw an error if `x` does not support reading metadata corresponding to dimension `dim`.
```
nalimilan, line 557: Can you use the same structure for the `colmetadata` and `dimmetadata` docstring, so that it's easier for users (and developers) to spot what's common and what's different? If necessary we can adapt the `colmetadata` docstring to improve it.

Instead of `collection[dim][key]` I would write `dimmetadata(x)[dim][key]`.

src/DataAPI.jl
```diff
@@ -515,6 +515,113 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+
+If `style=true` values are tuples of metadata value and metadata style. Metadata
+style is an additional information about the kind of metadata that is stored for
+the `key`.
+
+$STYLE_INFO
+
+The returned dictionary may be freshly allocated on each call to `dimmetadata` and
```
nalimilan, line 565:
```suggestion
The returned object may be freshly allocated on each call to `dimmetadata` and
```

src/DataAPI.jl
```diff
@@ -515,6 +515,113 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+is considered to be owned by `x` so it must not be modified.
+"""
+function dimmetadata(x::T, dim::Int; style::Bool=false) where {T}
+    if !dimmetadatasupport(T, dim).read
+        throw(ArgumentError("Objects of type $T do not support reading dimension metadata"))
+    end
+    return Dict(key => dimmetadata(x, dim, key, style=style) for key in dimmetadatakeys(x, dim))
+end
+function dimmetadata(x::T; style::Bool=false) where {T}
```
nalimilan, line 574: Why did you drop this check?
```suggestion
function dimmetadata(x::T; style::Bool=false) where {T}
    if !dimmetadatasupport(T).read
        throw(ArgumentError("Objects of type $T do not support reading dimension metadata"))
    end
```

src/DataAPI.jl
```diff
@@ -515,6 +515,113 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+pairs for all dimensions that have metadata.
+If `x` does not support dimension metadata return `()`.
+"""
+function dimmetadatakeys(x, dim::Int)
+    dimmetadatasupport(typeof(x), dim).read || return ()
+    throw(MethodError(dimmetadatakeys, (x, dim)))
+end
+function dimmetadatakeys(x)
+    Dict(dim => dimmetadatakeys(x, dim) for dim in 1:ndims(x))
+end
```
nalimilan, line 595: I don't remember the discussions offhand, but for `colmetadatakeys` we decided not to provide any fallback definition, in particular not to return `()` when metadata is not supported. Better be consistent for `dimmetadatakeys`, otherwise the API will be confusing (or, again, also adapt `colmetadatakeys` if necessary).

src/DataAPI.jl
```diff
@@ -515,6 +515,113 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+Throw an error if `x` does not support setting metadata for dimension `dim`.
+
+$STYLE_INFO
+"""
+function dimmetadata! end
+
+"""
+    deletedimmetadata!(x, dim::Int, key::AbstractString)
+
+Delete metadata for corresponding to dimension `dim` and key `key` of `x` and return `x`
```
nalimilan, line 611:
```suggestion
Delete metadata corresponding to dimension `dim` of `x` for key `key` and return `x`
```

src/DataAPI.jl
Tokazama: It still checks it for each dimensions because it dispatches to the above call.

src/DataAPI.jl
Tokazama: The reason I had the fall back is because the docs still say "If `x` does not support column metadata return `()`". Should the docs for both of these be changed then?

Tokazama:
@nalimilan, is this okay now?

src/DataAPI.jl
```diff
@@ -472,15 +472,17 @@ end
...
+`colmetadata(x, col, key)` returns a metadata value. Throw an error if `x` does not
 support reading column metadata or if `col` is not a column of `x`.
 
 `col` must have a type that is supported by table `x` for column indexing.
 Following the Tables.jl contract `Symbol` and `Int` are always allowed.
 
 If `col` is not passed return an iterator of `col => colmetadatakeys(x, col)`
 pairs for all columns that have metadata, where `col` are `Symbol`.
-If `x` does not support column metadata return `()`.
+
```
nalimilan, line 483: Let's keep this as is, as changing it would be breaking (https://github.com/JuliaData/DataAPI.jl/pull/62#discussion_r1294332342)
```suggestion
If `x` does not support column metadata return `()`.

```

src/DataAPI.jl
```diff
@@ -515,6 +517,109 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+"""
+    dimmetadatakeys(x, [dim::Int])
+
+If `dim` is passed return an iterator of metadata keys for which
+`dimmetadata(x, dim, key)` returns a metadata value. Throw an error if `x` does not
+support reading dimension metadata or if `dim` is not a dimension of `x`.
+
+If `dim` is not passed return an iterator of `dim => dimmetadatakeys(x, dim)`
+pairs for all dimensions that have metadata.
+
```
nalimilan, line 589: Let's do the same as for `colmetadatakeys`. Can you update the implementation accordingly?
```suggestion
If `x` does not support dimension metadata return `()`.

```

src/DataAPI.jl
nalimilan: That sentence is in the paragraph regarding the case when `col` is omitted. When it's passed, the docstring  says that we throw an error. We cannot change this for `colmetadatakeys` anyway as it would be breaking, so better make `dimmetadatakeys` consistent with that.

nalimilan (commented):
Sorry, looks like I missed your last push. I've made suggestions to ensure docstrings have the same structure as existing ones. There's also one decision to make about `dimmetadatasupport` (see [discussion](https://github.com/JuliaData/DataAPI.jl/pull/62#discussion_r1434854526)).

src/DataAPI.jl
```diff
@@ -515,6 +521,116 @@ Throw an error if `x` does not support metadata deletion for column `col`.
...
+is considered to be owned by `x` so it must not be modified.
+"""
+function dimmetadata(x::T, dim::Int; style::Bool=false) where {T}
+    if !dimmetadatasupport(T, dim).read
+        throw(ArgumentError("Objects of type $T do not support reading dimension metadata"))
+    end
+    return Dict(key => dimmetadata(x, dim, key, style=style) for key in dimmetadatakeys(x, dim))
+end
+function dimmetadata(x; style::Bool=false)
+    Tuple(dimmetadata(x, dim; style=style) for dim in 1:ndims(x))
```
quinnj, line 587: **[P2] Match the documented aggregate dictionary contract**

The docstring promises a dictionary containing dimensions with metadata, but this returns a tuple and visits every dimension. A valid 2D type that supports metadata only on dimension 2 can read `dimmetadata(x, 2)`, yet `dimmetadata(x)` throws at dimension 1, even when it defines the documented all-dimensions key iterator. The aggregate should follow that dictionary/subset contract, or the contract must be changed explicitly.

test/runtests.jl
```diff
@@ -353,6 +408,28 @@ end
     @test DataAPI.colmetadata!(tm, :col, "a", "100", style=:note) == tm
     DataAPI.emptycolmetadata!(tm)
     @test isempty(DataAPI.colmetadatakeys(tm))
+
+    @test isempty(DataAPI.dimmetadatakeys(tm, 1))
+    @test isempty(DataAPI.dimmetadatakeys(tm)[1])
```
quinnj, line 413: **[P2] Define and test the all-dimensions fixture iterator**

`TestMeta` only defines `dimmetadatakeys(x, dim)`, so this call raises `MethodError` and makes the full suite fail. The fixture needs the one-argument method, and this assertion should check the documented `dim => keys` iterator rather than assume an indexable collection of key sets.

test/runtests.jl
```diff
@@ -93,6 +104,39 @@ end
 
 DataAPI.emptycolmetadata!(x::TestMeta) = empty!(x.col)
 
+function DataAPI.dimmetadata(x::TestMeta, dim::Int, key::AbstractString; style::Bool=false)
+    return style ? x.dims[dim][key] : x.dims[dim][key][1]
+end
+
+function DataAPI.dimmetadata(x::TestMeta, dim::Int, key::AbstractString, default; style::Bool=false)
+    haskey(x.dims[dim], key) && return DataAPI.metadata(x, dim, key, style=style)
```
quinnj, line 112: **[P2] Use dimmetadata for present-key default lookup**

This test-fixture branch calls table-level `metadata(x, dim, key)`, which has no matching method. Once a dimension key exists, both ordinary and `style=true` calls with an explicit default throw instead of returning the stored value. Call `dimmetadata` here and cover present-key defaults; the current tests exercise only missing keys.

quinnj (commented):
[Reviewed by AI]

Reviewed exact head 6a0af432. On Julia 1.12.7, the PR suite has 205 passes and one error; current main has 177/177 passes. Three concrete findings follow: one aggregate API contract mismatch and two test-fixture defects. The existing whole-object trait design question remains open; this review does not resolve it.

PR #64, opened 2024-01-18
add a `groupby` method

rafaqz:
Add a `groupby` method to DataAPI so it can be shared accross packages like DataFrames.jl and DimensionalData.jl.

See: https://github.com/rafaqz/DimensionalData.jl/pull/591

### DIFF
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -528,8 +528,6 @@ function rownumber end
 Group `obj` into partitions specified by `args` and `kw`,
 which may vary according to the implementation for `obj`.
 
-The return value an iterable over the groups.
-
 To avoid type piracy and method ambiguities, implementations of `groupby` 
 must restrict the first argument to a type defined in the same package. 
 """
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -522,4 +522,14 @@ Return the row number of `row` in the source table.
 """
 function rownumber end
 
+"""
+    groupby(obj, args...; kw...)
+
+Group `obj` into partitions specified by `args` and `kw`,
+which may vary according to the implementation for `obj`.
+
+The return value an iterable over the groups.
+"""
+function groupby end
+
 end # module
--- a/src/DataAPI.jl
+++ b/src/DataAPI.jl
@@ -529,6 +529,9 @@ Group `obj` into partitions specified by `args` and `kw`,
 which may vary according to the implementation for `obj`.
 
 The return value an iterable over the groups.
+
+To avoid type piracy and method ambiguities, implementations of `groupby` 
+must restrict the first argument to a type defined in the same package. 
 """
 function groupby end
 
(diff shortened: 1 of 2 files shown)


src/DataAPI.jl
```diff
@@ -522,4 +522,14 @@ Return the row number of `row` in the source table.
 """
 function rownumber end
 
+"""
+    groupby(obj, args...; kw...)
+
+Group `obj` into partitions specified by `args` and `kw`,
+which may vary according to the implementation for `obj`.
+
+The return value an iterable over the groups.
```
bkamins, line 531: Why do we want to add this restriction to the contract? What if some package wants to e.g. return something that is indexable but not iterable?

src/DataAPI.jl
```diff
@@ -522,4 +522,14 @@ Return the row number of `row` in the source table.
 """
 function rownumber end
 
+"""
+    groupby(obj, args...; kw...)
+
+Group `obj` into partitions specified by `args` and `kw`,
```
bkamins, line 528: As commented in the original issue - I think we should add a comment that `obj` argument must ensure an appropriate type restriction so that there is no type piracy in the implementation of `groupby`.

bkamins:
Also a bump in package version is needed for the release.

src/DataAPI.jl
rafaqz: I assumed iterable is usually a subset of indexible, and a common property of the return value might make the interface more unified. But I can just remove this line.

src/DataAPI.jl
rafaqz: Done

bkamins (approved):

nalimilan:
Does it really make sense to have a docstring that says so little? Also, docstrings are exposed to users, so indications intended only at implementers would better go elsewhere. For join functions (`leftjoin` , etc.), we currently don't provide any docstring, and instead use a comment inside the code.

rafaqz:
Any objections to adding `aggregate` here at the same time for the same reason?

bkamins:
What packages use `aggregate` currently? DataFrames.jl does not use this name.

rafaqz:
Really? 
```julia
julia> aggregate
WARNING: both Rasters and DataFrames export "aggregate"; uses of it
 in module Main must be qualified
```

bkamins:
We will remove `aggregate` in the next release of DataFrames.jl then.

bkamins:
OK. @nalimilan - the state is that we export `by` and `aggregate`. I think no one uses them, we deprecated them many years ago.

Would you be OK to remove:
```
export by, aggregate

# TODO: remove definitions in 2.0 release
by(args...; kwargs...) = throw(ArgumentError("by function was removed from DataFrames.jl. " *
                                             "Use the `combine(groupby(...), ...)` or `combine(f, groupby(...))` instead."))
aggregate(args...; kwargs...) = throw(ArgumentError("aggregate function was removed from DataFrames.jl. " *
                                                    "Use the `combine` function instead."))
```
before 2.0 release (it is technically breaking, but in practice I think it is not problematic to drop it.

nalimilan:
Sure.

nalimilan (approved):

bkamins:
Released

